PR #3, opened 2022-09-16
NMSW-4 Setup GovUK Front End

jennikate:
# Ticket

NMSW-4

----

## AC

Add GovUK Frontend so we can use GDS components, classes, and styles

----

## To test

- run app locally
- go to localhost:3000
- > you should see a page with a h1, a button, and some paragraph text
- > they should be using the GDS Transport font
- > they should match the GDS style
- build app from terminal `npm run build`
- in terminal go to the dist folder `cd dist`
- serve the built version (production version)
- > you should see the same styles as the local version

----

## Notes

_I use `serve -s` for running the built version : https://www.npmjs.com/package/serve but you can use whichever package you like to serve built code for testing_


Adele-B (approved):
LGTM

PR #7, opened 2022-09-21
NMSW-8 GDS Header & Skip Link

jennikate:
# Ticket

NMSW-8

----

## AC

GIVEN The product site is up
WHEN I go to /
AND I can see the GDS styled header

GIVEN I am in the address bar
WHEN I tap tab once
THEN I am shown the skip link option
WHEN I click the skip link
AND I tap tab
THEN the header/navigation links have been skipped
AND the next interactive item is active (highlighted)

----

## To test

- run app locallly
- go to localhost:3000
- > you should see the header component at the top of the screen with the GOV UK logo and National Maritime Single Window as the service name

**Skip link**
- click on the address bar of your browser
- hit tab
- > you should see the skip link appear above the header and be highlighted in yellow (active state)
- hit tab again
- > you should see the GovUK logo highlighted (active state)
- click on the address bar again
- hit tab
- > you should see the skip link 
- click on the skip link
- hit tab
- > you should see the test button highlighted (it skips the header links)

**Responsiveness**
- scale the browser to mobile size 
- > you should see the header shrink but remain visible, as per GDS style

----

## Notes

- setup Jest and RTL
- added test for App.jsx that may be moved in future, tests that the header is applied to the page correctly
- added unit test for header text
- added unit test for skiplink element appearing, and having correct <a> references


Adele-B (approved):
LGTM

PR #8, opened 2022-09-21
NMSW-10 Add GDS Footer

jennikate:
# Ticket

NMSW-10

----

## AC

GIVEN The product site is up
WHEN I go to /
AND I can see the GDS styled footer

----

## To test

- run app locally
- go to localhost:3000
- > you should see the GDS footer with the Crown copyright logo
_the Crown copyright logo will not link to anywhere_

----

## Notes

The logo element will fail axe checks for colour contrast. It's a known situation https://github.com/alphagov/govuk-frontend/issues/2134

- add GDS footer
- comment out code for footer links as we don't need them now but will need them soon
- add background colour to body that matches footer background colour so if a page is shorter than the window it fills with the footer colour and not the content colour
- add background colour to root element (main div ID within body) that matches content background colour so that any page level components without a background colour specified will have that background, and anything specified (e.g header, footer) will use their specified colour
- added test for the text
- added test for the logo class name which will display the logo

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,3 +1,4 @@
+import Footer from './layout/Footer';
 import Header from './layout/Header';
 
 const App = () => {
@@ -15,6 +16,7 @@ const App = () => {
           <span>and a span where we forget to add the class, but should default to GDS font</span>
         </main>
       </div>
+      <Footer />
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


Adele-B (approved):
LGTM

PR #11, opened 2022-09-22
NMSW-11 Add Landing page

jennikate:
# Ticket

NMSW-11

----

## AC

GIVEN The product site is up
WHEN I go to /
AND I can see a page with a h1

----

## To test

- run app locally
- go to localhost:3000
  - you should see a page header with the GOVUK logo & a product title of National Maritime Single Window
  - neither of these will have links
  - you should see the GDS beta phase banner
  - you should see the GDS footer with the crown copyright logo
  - there will be no links in the footer
  - you should see a H1 saying National Maritime Single Window
  - you should see paragraph text saying Use this service to:

----

## Notes

- NMSW-47 created for when we have content for this page
- added a data-testid to the service name in the header so we can target that in tests to ensure it's rendering correctly when the service name is also shown on the page in other text
- added Landing page page component
- removed most of the test content from App.jsx
- as we don't have routing, rendering Landing directly in App.jsx for now

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,7 +1,10 @@
+// Page Layout Components
 import CookieBanner from './layout/CookieBanner';
 import Footer from './layout/Footer';
 import Header from './layout/Header';
 import PhaseBanner from './layout/PhaseBanner';
+// Pages
+import Landing from './pages/LandingPage/Landing';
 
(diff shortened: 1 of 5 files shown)


Adele-B (approved):
LGTM

PR #12, opened 2022-09-23
NMSW-13 Setup routing and add /sign-in

jennikate:
# Ticket

NMSW-13

----

## AC

GIVEN I am viewing the landing product page
WHEN I click on start now
THEN I am taken to the 'sign in' page

----

## To test

- run app locally
- go to localhost:3000
- > you should see a start now button
- click on start now button
- > you should be taken to /sign-in
- > you should see a h1 of sign in

----

## Notes

- added /sign-in page with a h1
- added tests and updated existing ones

Routing
- setup routing components
- moved internal URLs to <Link> elements and constants

Other things in PR
- updated Jest config to include all js/jsx files when calculating coverage
- set index.js to be excluded from coverage checking as it's the entry point
- removed invalid js import from index.html


Adele-B (approved):
LGTM

PR #13, opened 2022-09-26
NMSW-13 Remove rogue comma

jennikate:
# Ticket

NMSW-13

----

## FIX

Noticed an odd `,` on the page in the root element that shouldn't be there - removed it
Also identified it was resulting in aXe tools reporting that content on the page wasn't contained in a landmark

----

## To test

- run app locally
- go to localhost:3000
- scroll to the footer
- > there should be no white background below the footer
- > there should be no comma below the footer
- run aXe devtools
- > there should be no landmark issues

----

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,6 @@
       "quotes": [ "error",  "single"],
       "semi": ["error", "always"],
       "no-console": ["warn"],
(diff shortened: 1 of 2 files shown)


JayanthiiB (approved):
Great

PR #14, opened 2022-09-26
NMSW-23 Add accept/reject buttons to cookie banner

Adele-B:
# Ticket

NMSW-23

----

## AC

GIVEN I can see the cookie banner
WHEN I click 'do not accept'
THEN an essential cookie is set to store my preference to 'do not track'

GIVEN I can see the cookie banner
WHEN I click 'Accept'
THEN an essential cookie is set to store my preference to 'track'

----

## To test

- run app locally
- go to localhost:3000
- you should see the cookie banner at the top of the page
- it should have an accept button
- open dev tools
- you should not see a cookie for cookie preference
- > NOTE: if you have clicked a button in the cookie panel there will be a cookie, delete it before continuing
- click the accept button
- open dev tools
- view the cookie section
- > you should see a cookie for cookie preferences set to true
- refresh the page
- > you should see a cookie with preference set to true
- click the decline button
- open dev tools
- view the cookie section
- > you should see a cookie for cookie preferences set to false

----

## Notes

- You may have to manually click the refresh button (above the cookiePreference cookie) on the cookie section to see the preference update in dev tools

### DIFF
--- a/src/utils/findCookiePreference.js
+++ b/src/utils/findCookiePreference.js
@@ -1,7 +1,6 @@
 const findCookiePreference = (cookieName) => {
   const cookieArray = document.cookie.split(';');
 
-
   for (let i = 0; i < cookieArray.length; i++) {
     let cookiePair = cookieArray[i].split('=');
     if (cookieName == cookiePair[0].trim()) {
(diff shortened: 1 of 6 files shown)


src/utils/findCookiePreference.js
```diff
@@ -0,0 +1,17 @@
+const findCookiePreference = (cookieName) => {
```
jennikate, line 1: Wondering if we should call this something like 'cookieToFind' - just so on line 6 it's clearer what we're matching 🤔

jennikate (approved):
LGTM

PR #16, opened 2022-09-28
NMSW-14 Add Nav bar & user permitted state

jennikate:
# Ticket

NMSW-14

----

## AC

GIVEN I am viewing the sign in page
WHEN I sign in 
THEN I am taken to the 'dashboard'
AND I can see the navigation bar with dashboard, second page

----

## To test

- run app locally
- go to localhost:3000
- click on the start now button
- click on the sign in button (there is no authentication at this time)
- > you should be taken to a page with a h1 of signed in
- > you should see a navigation bar
- > you should see 'dashboard' highlighted in the navigation bar as the active page
- > you should see a 'second page' as an option in the navigation bar
- click second page
- > you should be taken to a page with a h1 of second page
- > you should see second page highlighted in the navigation bar as the active page (blue)
- > you should see dashboard as not highlighted (white)
- in devtools switch to mobile view
- > you should see a menu option in the top right
- click on menu
- > you should see the navigation options appear
- click on menu again
- > the navigation options should be hidden
- refresh the browser page 
- > you should be returned to the landing page /
- > you should no longer see the navigation bar

----

## Notes

Implemented
- userContext : stores user mocked logged in state
- useUserIsPermitted hook to check userContext (and in future session storage) and return true if user has permissions to access logged in content and false if not
- ProtectedRoutes : checks useUserIsPermitted and returns the route if user is permitted, or redirects to landing if user is not permitted
- Router : add 2x protected routes for testing purposes
- GDS Nav bar : including JS for menu show/hide and active state

Note implemented yet 
- there is no authorisation yet, the sign in is fake
- there is no persistence of logged in state on browser refresh, refresh will return to the logged out state, ticket NMSW-61


Adele-B (approved):
LGMT

PR #17, opened 2022-09-28
NMSW-19 Accessibility statement

Adele-B:
# Ticket

NMSW-19

----

## AC

GIVEN I am viewing a page
WHEN I click on the accessibility link in the footer
THEN I am shown the accessibility statement 

----

## To test

- run app locally
- go to localhost:3000
- go to the footer
- > you should see a link for Accessibility statement
- click that link
- > you should be taken to a page with a H1 of Accessibility statement for National Maritime Single Window
- there will be no other content on the page

----

### DIFF
--- a/src/pages/Footer/__tests__/AccessibilityStatement.test.jsx
+++ b/src/pages/Footer/__tests__/AccessibilityStatement.test.jsx
@@ -1,9 +1,9 @@
 import { render, screen } from '@testing-library/react';
 import AccessibilityStatement from '../AccessibilityStatement';
 
-describe('Cookie policy tests', () => {
+describe('Accessibility statement tests', () => {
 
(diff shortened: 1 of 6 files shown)


src/pages/Footer/__tests__/AccessibilityStatement.test.jsx
```diff
@@ -0,0 +1,10 @@
+import { render, screen } from '@testing-library/react';
+import AccessibilityStatement from '../AccessibilityStatement';
+
+describe('Cookie policy tests', () => {
```
jennikate, line 4: Guessing this should be Accessibility not Cookie ;)

jennikate (changes requested):
LGTM once we've changed 'cookie' to 'accessibility' in the accessibility statement test

PR #19, opened 2022-09-29
NMSW-17 Create Privacy Notice

Adele-B:
# Ticket

NMSW-17

----

## AC

GIVEN I am viewing a page
WHEN I click on the privacy link in the footer
THEN I am shown the privacy policy page

----

## To test

- run app locally
- go to localhost:3000
- go to the footer
- > you should see a link for Privacy Policy
- click that link
- > you should be taken to a page with a H1 of Privacy notice for National Maritime Single Window
- there will be no other content on the page

----


jennikate (approved):
LGTM

PR #23, opened 2022-10-03
NMSW-29 Display Form for Sign In page with email and password inputs and submit button

jennikate:
# Ticket

NMSW-29

----

## AC

GIVEN I am able to access the site
WHEN I go to /sign-in
THEN I can see a simple GDS styled form

----

## To test

- run app locally
- go to localhost:3000
- click on start now
- you should be taken to the sign in page (note there is no authentication)
- you should see a h1 of sign in
- you should see a form with an email input field and a password input field

----

## Notes

- there is no authentication, this form does nothing
- there is no validation at this time
- this ticket just places the inputs with their labels and hints on the page

This PR creates the basis for the form creator within our app and adds
- an input field that can take different types (email, password, text)
- a display form component that generates a form of fields along with labels, hints, action buttons, form data handler
- a sign in form with a false submit handler that just console logs the form data
- constants for form field types to ensure correct word is used when it's needed across components/functions


Adele-B (approved):
LGTM

PR #29, opened 2022-10-26
NMSW-120 Extend test coverage

jennikate:
# Ticket

Prep for NMSW-120

----

## AC

n/a

----

## To test

Run RTL tests

----

## Notes

Extending tests within form creator so we can use them as additional documentation on how to build forms with the creator

- removed some unneeded propTypes 
- made radioOptions required propTypes as we can't create radios without it
- added a formId for Cookie form as formId is now required (wasn't previously)
- added unit tests for InputRadio
- added integration tests for DisplayForm


Adele-B (approved):
LGTM

PR #48, opened 2022-11-10
NMSW-94 Persist most form data on page refresh (not passwords, not cookie preference)

jennikate:
# Ticket

NMSW-94

----

## AC

Given I am on a form
When I have entered anything into the form fields
And I refresh the page
Then the answers I have entered persist (they are not cleared)

Given I have refreshed the page and answers have persisted
When I submit the form
Then the fields that were persisted should be accepted and not error

Given I am on a form with a password field
When I have entered something in that field
And I refresh the page
Then the password field should become blank (we do not persist passwords)

----

## To test

- run app locally
- click sign in
- enter something in every field
- click refresh
- > email should remain
- > password should be cleared (not persisted)
- > sample field should remain
- click sign in button
- > password field should error
- enter password
- click sign in button
- > you should be taken to the dashboard
- go to second page form
- >  second page form form should be empty
- enter something in second page form form fields
- refresh page
- > values you entered should remain

----

## Notes
- currently nothing clears form fields, so anytime you go back to the form it will have values prefilled
- NMSW-160 starts the feature work to clear fields on submit, and other tickets will come for other scenarios

### DIFF
--- a/src/components/DisplayForm.jsx
+++ b/src/components/DisplayForm.jsx
@@ -1,5 +1,6 @@
 import { useContext, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
+import { FIELD_PASSWORD } from '../constants/AppConstants';
 import { UserContext } from '../context/userContext';
 import determineFieldType from './formFields/DetermineFieldType';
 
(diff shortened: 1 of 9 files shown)


Adele-B (approved):
LGTM

PR #49, opened 2022-11-11
NMSW-61 Persist sign in on refresh

Adele-B:
# Ticket

NMSW-61

----

## AC

Given I am logged in
When I refresh the page
Then I should remain logged in

Given I am logged in
When I close my browser tab
And I then reopen the nmsw app in a new tab
Then I should NOT be logged in

Given I am logged in
When I click sign out on the nav bar
Then I should be logged out and returned to the landing page

----

## To test

- run app locally
- go to localhost:3000
- click on the start now button
- click on sign in (mocked authentication still)
-> you should see the logged in nav
- refresh your browser
-> you should still be logged in and not redirected to landing page
- close your browser
- open a new browser tab
- go to localhost:3000
-> you should not see the nav (only visible if signed in)
- type localhost:3000/dashboard in the address bar
-> you should remain on the landing page, you cannot access /dashboard if you are not logged in
- click on start now
- sign in
- click on sign out in the nav
-> you should be taken to the landing page
- type localhost:3000/dashboard in the address bar
-> you should remain on the landing page

Added scenario to test
- go to second page form (nav item should be highlighted as second page) 
- fill in form and submit
- click return to homepage button
-> you should be taken to the dashboard
-> the Dashboard item should be highlighted in the nav)

----


src/layout/__tests__/Header.test.jsx
```diff
@@ -1,29 +1,30 @@
...
 describe('Header tests', () => {
 
-  it('should render the header with GovUK logo text', async () => {
-    await waitFor(() => { render(<MemoryRouter><Header /></MemoryRouter>); });
+  it('should render the header with GovUK logo text', () => {
+    render(<UserProvider><MemoryRouter><Header /></MemoryRouter></UserProvider>);
     expect(screen.getByText('GOV.UK')).toBeInTheDocument();
   });
 
   it('should render the service name text from the constants file', async () => {
```
jennikate, line 15: Something to remember to tidy up in next PRs -- we should remove the `async` if we have no `await`s in the test

jennikate (approved):
LGTM

PR #57, opened 2022-11-21
WIP: Conditional fields attached to radio buttons

Adele-B:
# Ticket

NMSW-140

----

## AC

Given I am on a form with conditional fields
When I select an option that has a connected text input
Then I should see the connected text input
 
Given I am on a form with conditional fields
When I select an option that does not have the connected text input
Then I should not see the connected text input
 
Given I can see the connected text input
When I click submit
And I have not entered anything in the connected field
And the connected field is required
Then I should see an error message for this connected field
 
Given a connected field has an error
When a click on the error link in the error summary for that field
Then the page should scroll to the question
And the connected field should be in focus
 
Given I can see an error for the connected field
When I enter something in the connected field
Then the error should clear
 
Given there are no other errors
When I click submit
Then the form should submit as normal

----

## To test

- run app locally
- go to localhost:3000
- sign in
- go to second page form
- select 'other' from the radio buttons
- > you should see a new, connected text input
- click submit
- > you should see an error that you must enter something in the connected text input
- the connected input should be highlighted red along with the rest of the question
- the connected input should be focused on (or the related radio button)
- enter something in the field
- click submit
- > you should see the confirmation page

----

### DIFF
--- a/src/components/DisplayForm.jsx
+++ b/src/components/DisplayForm.jsx
@@ -12,9 +12,32 @@ const DisplayForm = ({ fields, formId, formActions, handleSubmit }) => {
   const [errors, setErrors] = useState();
   const [fieldsWithValues, setFieldsWithValues] = useState();
   const [formData, setFormData] = useState({});
+  const [checkedOption, setCheckedOption] = useState();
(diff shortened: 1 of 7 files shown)


jennikate:
Replaced with #61 #62 #63 #64

PR #58, opened 2022-11-22
NMSW-33 Autocomplete input

jennikate:
# Ticket

NMSW-33

To make reviewing easier review & merger these two first:
https://github.com/UKHomeOffice/nmsw-ui/pull/59 - remove sample field from sign in
https://github.com/UKHomeOffice/nmsw-ui/pull/60 - scroll to h1 on errors not form id

Then rebase version-0.5.0 onto this PR and it will reduce it down to just adding the autocomplete input

----

## AC

Given I have to select from a predefined list of items
When I start to type an items name (After 2 characters)
Then I am shown a list of options that could match that item (From start of word)

----

## To test

UPDATE PACKAGES
- run `npm ci` to install alpha gov autocomplete package

NEXT

- run app locally
- go to second page
- submit the form
- > the country field should error as it is mandatory, the port field should not error as it is not mandatory

NEXT:

- type something in the country
- > you should see the filtered list of results
- select an item from the list
- type something in the port name (note only Dover and Portsmouth will work here)
- > you should see the filtered list of results that contain the port name & unlocode (if a unlocode exists)
- select an item from the list
- check the session storage
- > the field name & value should be in the form data object

NEXT:

- refresh the page
- > the field values should be persisted

NEXT:

- submit the form

----

## Notes

- we use a mock list of names to use for this component, this will be replaced with an API call in future
- the API will ultimately determine how the search works, whether it's on start of word, mid word, etc. So we will not test what is returned as part of this ticket

**ACCESSIBILITY**
We expect the following aXe failure for any autocomplete box on screen
```
Ensures all ARIA attributes have valid values : Invalid ARIA attribute value: aria-activedescendant="false"
```

there is an open PR to fix the aria-activedescendent issue: 
- https://github.com/alphagov/accessible-autocomplete/issues/434
- https://github.com/alphagov/accessible-autocomplete/pull/553/files

The aria-activedescendent looks to work correctly, if you use your mouse to select an item from the list then aria-activedescendent's value changes and voice over reads it out correctly
The error occurs because when the combobox pop up (list) is closed, the value of aria-activedescendent is set to = 'false' and false is an invalid value for it.
Some explanation of aria-activedescendant: https://www.holisticseo.digital/technical-seo/web-accessibility/aria-activedescendant/

**REACT DEPRECATED FEATURES**
An issue is open for warnings about deprecated lifecycles : https://github.com/alphagov/accessible-autocomplete/issues/418

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,14 +1,15 @@
 {
   "name": "nmsw-ui",
-  "version": "0.4.0",
+  "version": "0.5.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "nmsw-ui",
-      "version": "0.4.0",
+      "version": "0.5.0",
       "license": "MIT",
       "dependencies": {
+        "accessible-autocomplete": "^2.0.4",
         "govuk-frontend": "^4.3.1",
         "prop-types": "^15.8.1",
         "react": "^18.2.0",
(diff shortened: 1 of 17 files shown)


jennikate:
Replaced with #66 #67 #68 #69

PR #59, opened 2022-11-22
NMSW-33 Remove sample fields from sign in page

jennikate:
# Ticket

NMSW-33

----

## AC

Separating https://github.com/UKHomeOffice/nmsw-ui/pull/58 into multiple PRs for ease of review

This ticket removes the sample field we had to allow testing of some validation. We do not need this field on the sign in page.

----

## To test

- run app locally
- view sign in page
- there should only be email and password fields

----

## Notes

There is still no authentication


Adele-B (approved):
LGTM

PR #61, opened 2022-11-24
NMSW-140 Radio buttons with conditional text fields component (no errors or persisting values yet)

jennikate:
# Ticket

NMSW-140

----

## AC

Given I am on a form with conditional fields
When I select an option that has a connected text input
Then I should see the connected text input

Given I am on a form with conditional fields
When I select an option that does not have the connected text input
Then I should not see the connected text input

----

## To test

- run locally
- go to second page form
- > the last question should be a set of radio buttons
- select `dog`
- > the breed of dog question should be shown
- type something in that field
- select `cat`
- > the breed of dog field should be hidden
- > the breed of cat field should be displayed
- select `dog`
- > the breed of dog field should be shown
- > the breed of cat field should be hidden
- select `rabbit`
- > the breed of dog field should be hidden

----

## Notes

- there is no error handling yet
- there is no saving to session and prefilling from session yet

### DIFF
--- a/src/components/DisplayForm.jsx
+++ b/src/components/DisplayForm.jsx
@@ -14,7 +14,7 @@ const DisplayForm = ({ fields, formId, formActions, handleSubmit }) => {
   const [formData, setFormData] = useState({});
   const [sessionData, setSessionData] = useState(JSON.parse(sessionStorage.getItem('formData')));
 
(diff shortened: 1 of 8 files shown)


Adele-B (approved):
LGTM

PR #67, opened 2022-11-25
NMSW-33 Add basic autocomplete component

jennikate:
# Ticket

NMSW-33

----

## AC

Given I have to select from a predefined list of items
When I start to type an items name (After 2 characters)
Then I am shown a list of options that could match that item (From start of word)

----

## To test

- run app locally
- go to second page form
- go to country question
- type 2 letters of a country
- > you should see countries that match those letters in the list
- select a country
- go to the port question
- type `do` (we only have dover and portsmouth ports in the list right now)
- > you should see options for `Dover GB DOV` and `Dover Marina` and `Dover Other GB POR`
- select a port
- fill out rest of form
- click submit
- form should submit
- confirmation page reference should include the country and port you selected

----

## Notes

- there is no persist data on refresh yet
- there is no error handling yet
- we do not yet pass through country code or port code as separate fields in the dataset

### DIFF
--- a/src/assets/main.scss
+++ b/src/assets/main.scss
@@ -1,4 +1,5 @@
 @import '~govuk-frontend/govuk/all';
+@import "accessible-autocomplete";
 
 // ===================
 // Layout Overrides
--- a/src/components/__tests__/DisplayForm.test.jsx
+++ b/src/components/__tests__/DisplayForm.test.jsx
@@ -2,6 +2,7 @@ import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import DisplayForm from '../DisplayForm';
 import {
+  FIELD_AUTOCOMPLETE,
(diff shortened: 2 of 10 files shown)


jennikate:
Two issues found that will be resolved in next PR

- The autocomplete section in DetermineFieldType has an autocomplete='email' attribute - mistype?
- When you use esc key to exit the input it highlights the border pink - Is this okay?

Adele-B (approved):

jennikate:
Re field border being pink:

Pink will actually be whatever colour your browser is set to use as default.

It's happening because the escape key clears not only the combo box but the input focus class, however the input is still in focus so the browser default style shows.

If you look at the html for the input it has autocomplete__input--focused when you type
You hit escape it clears that class but the input is still in focus so browser focus style takes over

This is not something we can fix at this level, have raise an issue with alphagov: issue number 575

PR #69, opened 2022-11-25
NMSW-33 Extend autocomplete component to persist data on refresh and handle errors

jennikate:
# Ticket

NMSW-33

----

## AC

Given I have selected a value in an autocomplete field
When I refresh the page
Then the data should persist in the field

~~Given I have selected a value in an autocomplete field
And I have refreshed the page
When I submit the form
Then the full data should be submitted~~

Given I have attempted to submit a form with an autocomplete field
When the field is required
And has no value
Then an error should be shown

----

## To test

- run locally
- go to second page
- submit the form
- > the country field should error
- > the port field should error
- select a country
- refresh the page
- > the country field should remain populated
- select a port field with a unlocode
- refresh the page
- > the port field should persist with `name unlocode`
- submit the form
- > the form should submit

----

## Notes

This scenario is proving difficult to mock, and given we will only populate autocomplete fields from API calls I have left a comment to add the API function on persisting data as well as on populating fields for when we extend this component to get data from an API endpoint

### DIFF
--- a/src/components/formFields/InputAutocomplete.jsx
+++ b/src/components/formFields/InputAutocomplete.jsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import Autocomplete from 'accessible-autocomplete/react';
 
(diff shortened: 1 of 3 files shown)


jennikate:
Issues that will be addressed in next PR

- Errors for autocomplete do not clear when typed into, only when option is selected - is this expected?
- On refresh we (very briefly) see autocomplete values in the dropdown - will this be fixed if we add loading or something of the sort?

Adele-B (approved):

PR #71, opened 2022-12-05
NMSW-33 Move error summary above h1

jennikate:
# Ticket

NMSW-33

----

## AC

Given I have tried to submit a form with errors
When I view the error summary
Then I can see it ABOVE the h1

----

## To test

- submit a form without entering anything in the field
- > error summary should be above the h1

----

## Notes

also removed the `autocomplete="email"` from the autocomplete component as it's not a valid param and was put there by mistake

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -354,7 +354,7 @@ If you use a confirmation page
         }
       );
     } else {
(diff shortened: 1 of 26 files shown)


Adele-B (approved):
LGTM

PR #75, opened 2022-12-07
NMSW-205 Autocomplete not persisting values correctly

jennikate:
# Ticket

NMSW-205

----

## AC

Unable to replicate the value not persisting on local env - so needs to be tested on built version

Given I have typed at least two characters in an autocomplete field
When the list is displayed
Then I should see options related to what I typed

Given I can see a list of options
When I select an item from the list
Then it should show in the input field

Given I have a value in the input field
When I refresh the page
Then the value should persist

Given I have seen a value persist once
When I refresh the page again
Then the value should STILL persist

----

## To test

- make sure you're running docker
- `docker build -t nmsw-ui .`
- `docker run --name nmswui -p 8080:8080 nmsw-ui .`
- in browser go to localhost:8080

- type into the autocomplete fields
- select an item
- > check session data to see item including expandedDetails for the item exists
- refresh page
- > session data should ALL persist
- > value in input field should persist
- click on input
- > only the item that matches the input should show
- click on that item
- > check session data to see item including expandedDetails for the item exists
- refresh the page
- > session data should persist
- > input value should persist
- refresh page again
- > session data should persist
- > input value should persist
- complete form
- submit form
- > form should submit

----

## Notes

This solves the bug

HOWEVER
Once we attach the API's to get the data source this process will fail (as seen in the failing test)

AT that point, I think it makes more sense to do something like
- on page refresh
- if there is value for the autocomplete field in the session retrieve that and use it
- if there is no session value, but there is a value passed in from the form API use that
- otherwise field is empty

The general form creator will get the session value and pass it to the Input<x>, but for autocomplete we need to ignore that cause of the way it needs to retrieve and format the data we're better off just getting from the session directly I think

### DIFF
--- a/src/components/formFields/InputAutocomplete.jsx
+++ b/src/components/formFields/InputAutocomplete.jsx
@@ -85,8 +85,6 @@ const InputAutocomplete = ({ fieldDetails, handleChange }) => {
       const objectExpandedItem = objectItem[fieldDetails.fieldName];
       const objectAdditionalKey = objectExpandedItem[fieldDetails.additionalKey];
       const build = objectAdditionalKey ? `${objectExpandedItem[fieldDetails.responseKey]} ${objectAdditionalKey}` : objectExpandedItem[fieldDetails.responseKey];
-      // console.log('ak', build)
-      // const additionalKeyValue = objectItem[]
       valueToTest = build;
     } else {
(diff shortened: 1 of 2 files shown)


Adele-B (approved):
LGTM

PR #76, opened 2022-12-07
NMSW-117 Do not persist data after form cancel button click

Adele-B:
# Ticket

NMSW-117

----

## AC

Given I have cancelled a form that has values added
When I then return back to the form
Then the form is clear (i.e. answers do not persist after cancel)

----

## To test

- run app locally
- click sign in to go to second page
- > second page form should be empty
- enter something in form fields
- refresh page
- > values you entered should remain
- click cancel
- go back to second page form
- > form should be empty

----

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -49,8 +49,14 @@ Object structure:
 
 ```
 [action]: {
-  className: [required],
   label: [required],
+}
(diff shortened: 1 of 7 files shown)


jennikate (approved):
LGTM

PR #85, opened 2022-12-12
NMSW-155 Register account, password and confirmation pages

jennikate:
# Ticket

NMSW-155

----

## AC

GIVEN I have provided ALL of my details
WHEN I Submit my details
THEN I am asked to create my password

GIVEN I am on the password page
WHEN I click on the link "3 random words"
THEN I am taken to National Cyber Security Centre link in a new tab (ADD HERE)

----

## To test

- run app
- go to create account
- fill in email page and continue
- fill in your details page and continue
- > you should see create passwords page
- click the link to 3 random words
- > a new tab should open with the ncsc link
- click submit
- > both fields should error
- type no more than 3 characters in the password field
- click submit
- > that field should error as too short
- type something of at least 10 characters in the password field
- click submit
- > confirm password field should still error
- type something that does not match your password into the confirm password field
- click submit
- > confirm password field should still error
- type the same thing in both fields
- click submit
- > the confirmation page should load
- > the confirmation page should include the text `COMPANY NAME GOES HERE` and not the company name you submitted as we have not actually created the account yet this is coming in the next PR
- > the confirmation page should include a link to sign in
- click the sign in link

----

## Notes

- this still does not create a real account, that is coming in next PR when we POST to the API
- the sign in page is still a dummy page it does not do any authentication

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -109,23 +109,16 @@ Object structure:
 
 ```
 <action>: {
-  className: <required>,
   label: <required>,
 }
 ```
-If cancel button:
+
+If cancel button include the redirectUrl:
+
 ```
 <action>: {
   label: <required>,
   redirectURL: <required>,
-}
-```
-
-If cancel button:
-```
-[action]: {
-  label: [required],
-  redirectURL: [required],
 }
 ```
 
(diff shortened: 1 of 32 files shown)


jennikate:
new ticket number NMSW-224

PR #91, opened 2022-12-15
NMSW-155 All URL consts should include _URL

jennikate:
# Ticket

NMSW-155

----

## AC

Update const names to include _URL when they are a url

----

## To test

Review consts

----


jennikate:
new ticket NMSW-229

PR #92, opened 2022-12-15
NMSW-230 Add basic error page

jennikate:
# Ticket

NMSW-155

----

## AC

To enable testing add the base of an error handling page
And on any error redirect to this page with a message

This is needed so we can easily test what happens if the /registration endpoint returns anything other than a 200

----

## To test

- run app
- go to create account
- enter a new email address & repeat
- click continue
- > you should be taken to the your details page
- go back
- enter the same email address in the email fields
- click continue
- > you should be taken to the error page with a message that the email address already exists

----

## Notes

This is not the final implementation of any error handling, this is a basic form to allow testing.
Tickets are open to handle 4xx and 500 errors

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -9,6 +9,7 @@ import {
   ACCESSIBILITY_URL,
   COOKIE_URL,
   DASHBOARD_URL,
+  ERROR_URL,
   FORM_CONFIRMATION_URL,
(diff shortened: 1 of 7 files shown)


Adele-B (approved):
LGTM

jennikate:
ticket update to NMSW-230

PR #94, opened 2022-12-15
NMSW-232 Country phone code label

jennikate:
# Ticket

NMSW-155

----

## AC

Change `country code` to `country phone code` so that it's not confused with the country 2/3 digit codes

----

## To test

- run tests, all should pass
- review code

----


Adele-B (approved):
👍

jennikate:
updated ticket NMSW-232

PR #95, opened 2022-12-15
NMSW-233 Redirect if no email address when loading register your details page

jennikate:
# Ticket

NMSW-155

----

## AC

Given I have gone to `/create-account/your-details`
When there is no `email address` available in state
Then I should be redirected to the validate email page

----

## To test

- start a new browser session (new tab or new browser instance - this is to ensure there's nothing in your state)
- go to `http://localhost:3000/create-account/your-details`
- > you should be immediately redirected to `http://localhost:3000/create-account/email-verified`
- enter an email address (any, this is for testing only there is no validation here)
- click continue
- > you should be taken to your details

----

## Notes

We will change which page you're redirected to in future when the verify email process is built. For now this page allows testers to enter an email address and continue testing the registration flow

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -16,6 +16,7 @@ import {
   REGISTER_ACCOUNT_URL,
   REGISTER_CONFIRMATION_URL,
   REGISTER_EMAIL_URL,
+  REGISTER_EMAIL_VERIFIED_URL,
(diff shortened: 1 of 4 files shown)


Adele-B (approved):
👍

jennikate:
Ticket changed to NMSW-233

PR #96, opened 2022-12-15
NMSW-234 PATCH registration

jennikate:
# Ticket

NMSW-155

----

## AC

Given I have clicked submit on the create-account/your-password page
When I have no errors
Then the code sends a PATCH request to /registration

Given a PATCH request has been sent to /registration
When a success (200) response is received
Then I am taken to the confirmation page

Given a PATCH request has been sent to /registration
When an error response is received
Then I am redirected to the error page

Given a PATCH request has been sent to /registration
When no valid response is received
Then I am redirected to the error page
(this scenario is not tested in the UI, it's the `catch` scenario of the try/catch for if a promise fails unexpectedly)

----

## To test

- run BE in docker
- run app
- go to `http://localhost:3000/create-account/email-verified`
- enter an email address you have NOT previously registered
- fill out your details
- fill out password
- click submit
- > you should be taken to the error page with 'user not registered' message
- click continue to be taken back to the email page
- enter an email address you HAVE previously registered
- continue through the form and submit
----

## Notes

Handling errors is not fully implemented and so sessions are not cleared when errors occur in the test case above.
This is known and will be handled when we finish error implementation for 4xx and 500 errors in future tickets

### DIFF
--- a/src/hooks/usePatchData.js
+++ b/src/hooks/usePatchData.js
@@ -1,0 +1,21 @@
+import axios from 'axios';
+import Auth from '../utils/Auth';
+
+const usePatchData = ({ url, dataToSubmit, pageSource }) => {
+  const source = pageSource || location.pathname.substring(1);
+  const data = axios.patch(url, dataToSubmit, {
+    headers: { Authorization: `Bearer ${Auth.retrieveToken()}` },
+  })
+    .then((resp) => { return resp; })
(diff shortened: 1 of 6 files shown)


jennikate:
ticket updated to NMSW-234

Adele-B (approved):
LGTM

PR #97, opened 2022-12-15
NMSW-235 Registration confirmation page

jennikate:
# Ticket

NMSW-155

----

## AC

Given I have clicked submit on the password page
When there are no errors 
And my account has been created (200 response from API)
Then I am shown the confirmation page

----

## To test

- run BE in docker
- run app
- go through the create account process and submit
- > you should be shown the confirmation page
- > it should have <your company name> is created in the panel

----


jennikate:
new ticket NMSW-235

Adele-B (approved):
LGTM

PR #99, opened 2022-12-16
NMSW-226 Do not allow spaces in password creation fields

jennikate:
# Ticket

NMSW -226

----

## AC

Given I click submit
When there are spaces in the password value
Then I am shown an error

----

## To test

- run app
- go to /create-account/your-password
- click submit
- > `Enter a password` error should be displayed
- enter a password that has spaces in it but is less than 10 characters
- click submit
- > `Password must be at least 10 characters` error should be displayed
- enter a password that is 10 spaces only
- click submit
- `Enter a password` error should be displayed
- enter a password that is at least 10 character and has no spaces
- enter confirmation password as the same
- click submit
- > no errors should show
- > if you have the BE running in docker you should be taken to the confirmation page

----

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -26,7 +26,9 @@ Validating fields
 - [Conditional](#conditional)
 - [Email Format](#email-format)
 - [Match](#match)
+- [Maximum Length](#maximum-length)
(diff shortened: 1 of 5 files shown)


Adele-B (approved):
LGTM

PR #104, opened 2022-12-22
NMSW-219 Add create account link to sign in page

jennikate:
# Ticket

NMSW-219

----

## AC

Given I click on the create one now link
Then I am taken to the /create-account/email-address page 

----

## To test

- go to sign in page
- there should be a create one now link
- click the link
- you should be taken to /create-account/email-address

----

### DIFF
--- a/src/assets/main.scss
+++ b/src/assets/main.scss
@@ -40,9 +40,15 @@ body {
   }
 }
 
+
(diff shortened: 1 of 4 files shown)


Adele-B (approved):
LGTM

PR #105, opened 2022-12-22
Test registration feature

JayanthiiB:
1. **Test changes made  in the cookie feature** 

- Changed the save button locator
- changed the scenario in the cookie feature as

> - when the user is on the cookie preference page by clicking view cookies 
> - user clicks accept/rejects cookie on the cookie banner
> - user is shown with confirmation banner
> - when clicking change your cookie settings on the confirmation banner, auto closes the confirmation banner so the Hide cookie message won't be shown to hide the confirmation banner
> 

2.    **Added test to successful user registration -positive scenario**

> ## To test
> To test

run npm install to add faker dependency
```
npx cypress open 
```

- then  choose cookie.feature to run cookie feature specs
- choose registration.feature to run registration feature specs


jennikate (approved):
LGTM

PR #108, opened 2022-12-23
NMSW-178 Error/Account exists page

jennikate:
# Ticket

NMSW-178

----

## AC

**Scenario: Submit email address for existing account**
GIVEN I have completed my email address on the create account/email address page
AND the email address HAS already been used for an active account
AND a verification request is NOT currently active
WHEN I submit it
THEN an error page is shown with a "400: User is already registered" message
AND a link to the sign-in page

**ACTIONS:**
Given the app has requested account creation
When an error response is received
Then I am shown the Error page with the relevant error message and link

----

## To test

- run app
- go to create account
- enter an email address for an existing account
- > you should be taken to account already active page
- > your email address should be in text
- > there should be a sign in link
- click sign in
- go to a new tab
- go directly to `http://localhost:3000/create-account/account-already-exists`
- > account already active page should load
- > but you will NOT have your email address on page

----

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -10,6 +10,7 @@ import {
   COOKIE_URL,
   DASHBOARD_URL,
   ERROR_URL,
+  ERROR_ACCOUNT_ALREADY_ACTIVE_URL,
   FORM_CONFIRMATION_URL,
   LANDING_URL,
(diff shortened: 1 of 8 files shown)


Adele-B (approved):
LGTM

PR #109, opened 2022-12-23
Test register user throws error if user already exist

JayanthiiB:
Test to validate registration error message if user already exist

----

## To test
To test
To run just the validation feature(Should not register existing user) , pls comment other scenarios
```
npx cypress open 
```
choose registration.feature to run registration feature specs


jennikate (approved):
👍

PR #113, opened 2023-01-03
NMSW-264 Add case sensitive and non-case sensitive 'fields must match' validators

jennikate:
# Ticket

NMSW-264

----

## AC

BUG - email address matching fails when casing is different

----

## To test

- go to create account
- enter `testemail@email.com` for email
- enter `boo@email.com` for second email
- click submit
- > field should error due to mismatch
- enter `TESTemail@EMAIL.COM` for second email
- click submit
- > field should not error

----

## Notes

I split out the error scenarios for DisplayForm testing into their own fields to make them easier to manage as part of this ticket


Adele-B (approved):
LGTM

PR #115, opened 2023-01-05
NMSW-283 Setup html in <noscript> for if user has javascript disabled

jennikate:
# Ticket

NMSW-283

----

## AC

Handle no javascript with a styled page

----

## To test

- run app locally
- turn off javascript in browser
- refresh page
- > the 'no javascript' page should load
- turn javascript back on
- refresh page
- > the normal app should load

----

## Notes

We don't have content for the page (future ticket), this ticket was a spike but as the style code is working I figure we might as well merge that part.


Adele-B (approved):
LGTM

PR #117, opened 2023-01-05
Version 0.6.1

jennikate:
Merge Version 0.6.1


Adele-B:
- Cypress tests are failing due to lack of token from BE
- registration.steps.js cypress file is missing a closure `})` on line 134
- In the cypress email registration process `Continue` button needs to be updated to `Send confirmation email`

jennikate:
This is not a release that will go live so decided Cypress bugs can be fixed as part of the next release as the Cypress files to be changed will still fail on this release due to a change in the API we connect to for verifying emails that is on dev now, but that we haven't updated the Front End to handle yet.

Adele-B (approved):
👍

PR #118, opened 2023-01-06
Change class to className so js css is applied properly

jennikate:
# Ticket

----

## AC

Hidden fields component is showing fields when they should be hidden

>> discovered it was due to an error in having `class` instead of `className`

----

## To test

- run app
- open second page
- > conditional fields should be hidden 
- click on cat
- > conditional field should show

----


Adele-B (approved):
LGTM

PR #120, opened 2023-01-09
NMSW-167 Enable redirect via sign in

jennikate:
# Ticket

NMSW-167

----

## AC

Enabler for redirecting a user via the sign-in page if a token expires/doesn't exist/is invalid

Given a user successfully submits in on the sign in page
When there is a redirectURL in the state
Then the URL in the redirectURL value is loaded

Given a user successfully submits in on the sign in page
When there is NO redirectURL in the state
Then the dashboard is loaded

Given a user has been redirected to sign in from a form page
And they had filled in some of the fields on that page
When they successfully sign in
And are returned to the form page
Then the values of the fields they filled in are persisted

----

## To test

- run app
- sign in
- go to second page
- fill in some files
- click the 'redirect to sign in' test button
- sign in (note this is still a fake sign in)
- > you should be taken back to the second page
- > the values you entered should be persisted

----

## Notes

While building this and looking at how we handle session data for forms and sign in in particular I've decided not to store any sign in form data in the session. We already did not store password and there's no requirement to store email here. User can still fill the email value from their browser saved usernames.

Added a form type for sign in so we can treat it different to other forms.

### DIFF
--- a/src/App.test.jsx
+++ b/src/App.test.jsx
@@ -111,11 +111,11 @@ describe('App tests', () => {
   it('should clear formData when a footer item is clicked', async () => {
     const user = userEvent.setup();
     render(<MemoryRouter><App /></MemoryRouter>);
-    const startButton = screen.getByRole('button', { name: 'Start now' });
+    const startButton = screen.getByRole('link', { name: 'create an account' });
     await user.click(startButton);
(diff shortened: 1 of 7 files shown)


Adele-B (approved):
LGTM

PR #122, opened 2023-01-09
NMSW-167 Move API calls within functions

jennikate:
# Ticket

NMSW-167

----

## AC

Move POST and PATCH to within the handleSubmit functions of the pages that use them

----

## To test

- run app
- go to create account
- add new email address and continue
- > you should be taken to the 'verif your email page'
- go back to create account
- add the same email address again and continue
- > you should be taken to the 'user already exists' page
- go back to create account
- add a new email address and continue
- click on the 'verify email' link to bypass verification
- > you should be taken to the enter your details page
- enter your details and continue
- enter your password and continue
- > you should be taken to the 'verification link expired' page [we only have a 401 invalid token status returned at this time]

----

## Notes

- enables greater control over how to handle various errors
-  reduces complexity of passing many props to a slightly more generic hook to achieve the same without saving much code


Adele-B (approved):
LGTM

PR #123, opened 2023-01-09
Fix page titles

jennikate:
# Ticket

----

## AC

Page titles should mimic the H1

----

## To test

----

## Notes

FIxed the page title so they show the H1 where available, and default back to service name where not


Adele-B (approved):
👍

PR #124, opened 2023-01-09
NMSW-166 Handle axios or 500 error responses

jennikate:
# Ticket

NMSW-166

----

## AC

Handle 500 errors

----

## To test

Review tests
run unit/integration tests
they should pass

----

## Notes
- added handling for if axios itself throws an error
- made the generic fallback error handling not require a message so it can handle 500 responses


Adele-B (approved):
LGTM

PR #126, opened 2023-01-10
NMSW-180 Rename page & component

jennikate:
# Ticket

NMSW-180

----

## AC

Rename 'error' 'generic error' to 'messages' as we want to use the same component for non-error messages

----

## To test

Run unit/integration tests

----

## Notes

- moved all error and confirmation pages into a 'messages' folder
- updated GenericError page to GenericMessage
- we can now use that for any basic message be it an error or other that needs just `h1`, `optional paragraph`, `link to next url to visit`
- and any specific message pages such as a form confirmation, or the account active and the like that need different styles or content we can store their components in this folder too


Adele-B (approved):
LGTM

PR #127, opened 2023-01-10
NMSW-278 Voyage details page (no voyages)

Adele-B:
# Ticket

NMSW-278

----

## AC

Given I have logged in
I am landed on the /your-voyages page
With an option to 'start a new report' (button will not work)

----

## To test

- run app
- sign in
- > you should be taken to /your-voyages page
- Go though the code and check that all instances of Dashboard are changed to Your voyages

----


jennikate (approved):
LGTM

PR #128, opened 2023-01-10
Add a wait in tests to prevent timing errors

jennikate:
----

## Notes

We occasionally had 2 tests failing inexplicably and randomly - they would fail when we'd not touched related code, and if you run the test suite again then they would pass.

Investigating it looks like the `expect` statement that checked for error text to have appeared on the page was running before the errors has been generated.

To try to prevent this we've added waits around the error summary title to wait for it to appear before running the rest of the assertions


Adele-B (approved):
👍

PR #130, opened 2023-01-11
Add cypress environment variables

JayanthiiB:
Added configuration files for different test environments

Added scripts to run test using config files

## To test

To run all tests in dev environment
```
npm run cypress:run:dev
```
To open the test in cypress runner to run particular test
```
npm run cypress:open:dev
```
and select the feature file to test


jennikate (approved):
Discussed the // in the dev url with Jayanthi -it's fixed in the next PR so signing off this one on the basis that fix is comiing

JayanthiiB:
Changed the URL

PR #131, opened 2023-01-12
Update linting

jennikate:
Adding stricter linting rules, and updating existing files, to help us with consistent formatting and avoid potential issues.
Doing now before we get too many fully functioning features in place!


Adele-B (approved):
👍

PR #135, opened 2023-01-13
NMSW-180 Verify email address - success path

jennikate:
# Ticket

NMSW-180

----

## AC

Given a user clicks their 'verify your email address' link in their email
When the FE receives that request
Then the FE sends a GET request to /check-token endpoint with the token details from the link

Given the FE app has sent a  GET request to /check-token endpoint
When the response is 204
Then the email address verified page is loaded with a continue button

Given the user has been shown the email address verified page
When they click the continue button
Then they are taken to the /your-details page 

Given they are taken to the /your-details page
When they complete the registration form and submit
Then their account is activated

----

## To test

-   // to test this as a user in browser
-   // make new email address
-   // add email to GovNotify
-   // create account using that email
-   // get link from email - that token lasts 8 hours

_let me know if you need to me to make it, I can make for jentestingemailsforwork+<todaysdate>@gmail.com & we can use that url_

- go to /create-account/email-address
- add new email address and submit
- > you should be taken to confirm your email page
- > you should have a verification email
- click the verification email link
- > you should be taken to the email has been verified page
- click continue
- > you should be taken to /your-details page
- complete the your details and password pages and submit
- > you should be taken to the account created page

----

## Notes

Have to test against Dev BE for now as code is on the release branch but release not moved to main yet as other things to continue testing with it before it is merged

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,7 +28,7 @@
   "rules": {
     "eol-last": ["error", "always"],
     "import/prefer-default-export": ["off"],
-    "max-len": ["error", { "code": 250, "ignoreStrings": true, "ignorePattern": "^path " }],
+    "max-len": ["error", { "code": 250, "ignoreStrings": true, "ignoreComments": true, "ignorePattern": "^path " }],
     "no-plusplus": ["error", { "allowForLoopAfterthoughts": true }],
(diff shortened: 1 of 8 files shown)


Adele-B (approved):
LGTM

PR #136, opened 2023-01-16
NMSW-181 Verify email - account already exists

jennikate:
# Ticket

NMSW-181

----

## AC

Given a user clicks their 'verify your email address' link
When the FE receives that request
Then the FE sends a GET request to /check-token endpoint with the token details from the link

Given the FE app has sent a  GET request to /check-token endpoint
When the response is 401: "Token was already used"
Then we show the user the 'error' page with 'You already have an account' message and sign in link

Given the user is on the error page
When the error reason is 'You already have an account'
And the user clicks the button
Then they're taken to the /sign-in page

----

## To test

- go to /create-account/email-address
- add new email address and submit
- you should be taken to confirm your email page
- > you should have a verification email
- click the verification email link
- > you should be taken to the email has been verified page
- click continue
- > you should be taken to /your-details page
- complete the your details and password pages and submit
- > you should be taken to the account created page
- go back to your email
- click the verification email link again
- > you should be taken to 'you already have an account' page
- click the sign in button

----

## Notes
Confirmed with Endi that `401: "Token was already used"` is the indicator that the account has been registered

### DIFF
--- a/src/constants/AppAPIConstants.js
+++ b/src/constants/AppAPIConstants.js
@@ -7,4 +7,5 @@ export const REGISTER_CHECK_TOKEN_ENDPOINT = `${apiUrl}/check-token`;
 // Responses
 export const AXIOS_ERROR = 'Network Error';
 export const TOKEN_INVALID = 'Token is invalid or it has expired';
+export const TOKEN_USED_TO_REGISTER = 'Token was already used';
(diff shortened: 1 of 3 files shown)


Adele-B (approved):
LGMT

PR #138, opened 2023-01-16
NMSW-168 Add a details input field

jennikate:
# Ticket

NMSW-168

----

## AC

The ticket requires an email input field contained within a details component. 
This PR creates that in our form creator

----

## To test

- run app & login
- open second page
- > the first form field should be a `<details>` link
- click the link
- > the email address field should be displayed
- submit the form
- > it should error as normal
- complete the form and then submit
- > it should continue to the confirmation page as normal

----

## Notes

If you do not expand the details field and click submit it will still error as the validation is set to be required for this form.

To enable this, I've had to change the ternary of grouped vs single to be an option of grouped, details, or single
To prevent existing things breaking I've also set the default to always be `single` so if no display type is passed in the form will render the field as a `single`  component

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -212,7 +212,7 @@ Object structure
   type: FIELD_RADIO,
   className: <required>,
   fieldName: <required>,
-  grouped: true,
+  displayType: DISPLAY_GROUPED,
   hint: <optional>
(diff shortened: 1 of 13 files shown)


Adele-B (approved):
LGTM

PR #139, opened 2023-01-17
NMSW-318 File storage and download

Adele-B:
# Ticket

NMSW-318

----

## AC

Given  a user clicks on the templates tab in the nav
When they are taken to the /templates page
Then they see links to download the templates

Given a user is on the /templates page
When they click a template link
Then the file is downloaded
And it can be opened without errors

----

## To test

- Run the app locally
- Go to localhost:3000
- Sign in
- Click on templates in the nav
- > You should be taken to /templates
- Click on one of the template links
- > The file should start to download
- Open the file you downloaded
- > It should open without error

----

## Notes

The files here will not be the final files - these will be the old templates until the new ones are ready


jennikate (approved):
Looks good!

PR #140, opened 2023-01-17
NMSW-322 Display and use defaultValues when exist and no sessionData exists

jennikate:
# Ticket

NMSW-322

----

## AC

Given a field has been assigned a defaultValue (in the formFields object either directly or via variable)
When the form is rendered
And there is no session data value for that field
Then the field's value should be set to the defaultValue
And the formData object should include that field with the defaultValue

Given a field has been assigned a defaultValue (in the formFields object either directly or via variable)
When the form is rendered
And there IS session data value for that field
Then the field's value should be set to the sessionData value
And the formData object should include that field with the sessionData value

----

## To test

- go to SecondPage
- > the`First name missing display type test` field should have a sample value in it
- click submit
- > the field should not error
- type a new value in the field
- refresh the page
- > the new value should persist

----

## Notes

Requirement: when a default value is assigned to a field it is should be shown on render

Issue: it wasn't being shown due to a line of code that was setting the value to be sessionData.value or null

Fix: change that line to show sessionData.value or defaultValue, where defaultValue will be either a value or undefined on the fieldData passed to DisplayForm


Adele-B (approved):
LGTM

PR #141, opened 2023-01-17
Add TEMPLATE_PAGE_URL to TOP_LEVEL_PAGES

Adele-B:
## Notes

Adding TEMPLATE_PAGE_URL to TOP_LEVEL_PAGES array so that formData clears on page and back button is removed from the page


jennikate (approved):
👍

PR #146, opened 2023-01-18
NMSW-218 Resend verification email if user attempts to register with an email address already in awaiting verification state

jennikate:
# Ticket

NMSW-218

----

## AC

**DESIGN:**

use existing confirmation page

**ACTIONS:**

Given the app has sent a POST request to /registration
When the response is that we have their email address but they are not active yet
Then the app should send a POST request for that email to /resend-verification-email
And show the user the 'Check your email' verification confirmation screen

----

## To test

- go to `/create-account/email-address`
- enter a new email address & submit
- > you should see the check your email page (`/create-account/check-your-email`)
- > you should receive a verification email
- return to `/create-account/email-address`
- enter the same email address
- > you should see the check your email page (`/create-account/check-your-email`)
- > you should receive a new verification email
- click on the link in the email
- > you should be able to complete registration

----


Adele-B (approved):
LGTM

PR #147, opened 2023-01-18
NMSW-323 Create loading spinner component

Adele-B:
# Ticket

NMSW-323

----

## AC

Given a user has slow internet
When they action something that needs to process data
Then a loading spinner is displayed until the data is processed

----

## Notes

Currently there is no implementation of the spinner but this code creates the spinner which will be implemented where there is asynchronous code in future tickets


jennikate (approved):
👍

PR #149, opened 2023-01-19
NSMW-324 Remove test pages (second page)

jennikate:
# Ticket

NMSW-324 prework

----

## AC

Remove the page we were using to test form elements as we no longer need it

----

## To test

- run tests, all should pass
- run app
- > there should be no Second page in the nav
- > nav items should work

----

## Notes

Also did a bit of tidying up of the order and labelling of urls/page routes


Adele-B (approved):
LGTM

PR #161, opened 2023-01-24
Modify test with delete endpoint

JayanthiiB:
Modified tests to work to reuse the email with the DELETE user endpoint 

## To test

```
npm run cypress:open:dev
```


jennikate (approved):
👍

PR #162, opened 2023-01-25
NMSW-216 Auth token missing or invalid

Adele-B:
# Ticket

NMSW-216

----

## AC

Given the system is trying to hit an endpoint
When the auth token is missing or invalid
Then take the user to sign-in page
 
Given the FE is trying to hit an endpoint that needs a token
When there is not token available
Then the FE redirects the user to the login page

----

## To test

- Review the tests and run them
- > They should pass

----

## Notes

- Currently we do not have forms we cannot test forms being pre-filled.
- We don't have an API endpoint ready to test the functionality so we have mocked the api response in tests
- The code in the browsers will not work and will need to be refactored once the BE is in place
- This will need to be implemented on each handleSubmit for the file uploads once each section is built


jennikate (approved):
👍

PR #165, opened 2023-01-27
NMSW-354 Your voyages page skeleton page

Adele-B:
# Ticket

NMSW-354

----

## To test

- Run app
- Go to localhost:3000
- Sign in
- > You should be taken to the /your-voyages page and see mocked voyages
- Compare with the prototype
- > It should match
- Switch to mobile view
- It should match the video provided by UX designer

----

## Notes

- The filter buttons do nothing at the moment
- The voyage data is hardcoded and the links will not work


jennikate (approved):
LGTM

PR #167, opened 2023-01-30
NMSW-367 Make templates available for download [Voyage details, crew, passenger]

Adele-B:
# Ticket

NMSW-367

----

## AC

As an agent
I want to download the latest templates
So I can submit them in the right format

Given I have clicked on the FAL 1 template
When it downloads
I have an xlsx file with the correct field labels and filename

Given I have clicked on the FAL 5 template
When it downloads
I have an xlsx file with the correct field labels and filename

Given I have clicked on the FAL 6 template
When it downloads
I have an xlsx file with the correct field labels and filename

----

## To test

- Run the app
- Go to localhost:3000
- Sign in
- Go to /templates
- Download the files
- > They should download and open correctly AND be the correct templates for MVP

----

## Notes

- Templates are in ticket NMSW-279
- It is okay if the sizes of the files after downloading do not match the size shown on the screen as users just need to see the file size to have an idea of how big it is
- There is not an easy way to test DownloadFile.js file - this will be covered in end-to-end testing


jennikate (approved):
LGTM

PR #168, opened 2023-01-30
NMSW-360 Fix sign in error bug

Adele-B:
# Ticket

NMSW-360

----

## To test

- Run the app
- Go to `localhost:300/sign-in`
- Try to sign in with a valid email but an incorrect password
- > The page should scroll to the top on error

----


jennikate (approved):
lgtm

PR #169, opened 2023-01-31
NMSW-356 Disable submit or show loading spinner when processing

jennikate:
# Ticket

NMSW-356

----

## AC

Given a submit button has been clicked
And there are no errors (aka form is submitting)
Then the submit button should become disabled

_When a component is waiting on an API call response
Then a loading spinner should show_

The scenario we currently have to test this is the verify email scenario
And the loading spinner only shows when the site is slow

----

## To test
- in dev tools, network tab, set your speed to slow 3G
- run app
- go to `/create-account/email-address`
- click `send confirmation email` button
- > page should show errors
- > `send confirmation email` button should remain enabled
- enter valid email addresses
- click `send confirmation email` button
- > you should see the button change to disabled state before the submit completes and the next page loads
- get your link to verify your email address
- paste the link into the address bar where you have slow 3G loading
- > you should see a loading spinner on the page before the page content loads

----

### DIFF
--- a/src/components/DisplayForm.jsx
+++ b/src/components/DisplayForm.jsx
@@ -17,7 +17,7 @@ import { scrollToTop } from '../utils/ScrollToElement';
 import Validator from '../utils/Validator';
 
 const DisplayForm = ({
-  fields, formId, formActions, formType, pageHeading, handleSubmit, children, removeApiErrors,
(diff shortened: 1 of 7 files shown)


Adele-B (approved):
LGTM

PR #166, opened 2023-01-30
NMSW-375 Add webpack bundling

jennikate:
# Ticket

NMSW-375

----

## AC

Improve build bundle sizes

----

## To test

- delete your dist folder (if you have one)
 run npm run build
- in package.json change start to be `"start": "webpack serve --mode production",` (so it runs the production mode)
- run npm start

----

## Notes

1 - increase size from 244 to 512
2 - add webpack standard chunking
3 - bundle Templates component separately as it's large and rarely changes

PR #170, opened 2023-01-31
Webpack: revert optimisation changes from PR166

jennikate:
Revert https://github.com/UKHomeOffice/nmsw-ui/pull/166 changes as they've caused FOUC (flash of unstyled content) issues and we need to find a new fix


Adele-B (approved):
👍

PR #174, opened 2023-02-02
NMSW-257 Sign out

Adele-B:
# Ticket

NMSW-257

----

## AC

As a signed in user
I want to sign out
So my session finishes
 
Given a user has an active authentication token
Then there is a sign out option on the nav bar

Given a user clicks on sign out
Then their session data is cleared (including their auth token)
And a POST is made to /sign-out to blacklist the token
And they are taken to the landing page

----

## To test

- sign in
- > you should have a sign out option in the nav
- open dev tools, network tab
- click sign out
- > you should have a POST request to the /sign-out endpoint
- open dev tools and check the session
- > there should be no session data for NMSW

----


jennikate (approved):
LGTM

PR #175, opened 2023-02-02
Configure multiple html reporter

JayanthiiB:
Configured Multiple-cucumber-html-reporter 

## To test

First run E2E test by, will run E2E test and generate basic cucumber html report
```
npm run cypress:run:dev
```
and 
```
npm run cypress:report
```
to generate multiple-html-report under reports folder named index.html


Adele-B (approved):
👍

PR #181, opened 2023-02-06
NMSW-386 Change your password skeleton page

Adele-B:
# Ticket

NMSW-386

----

## AC

No functionality but can put together page layout & render tests

----

## To test

- Run app
- Go to localhost:3000
- Sign in
- Click on Your details in the nav
- Click on 'Change your password'
- > You should be taken to a page with a blank form to change your password 
- > It should match the prototype (with text correctly in the middle of text inputs)
- Click continue
-> All the fields should error
- Type a password less than 10 characters into the new password field
- > There should be a length error
- Type a password that does not match in the New password and Confirm new password fields
- You should get a passwords do not match error
- Enter passwords in the correct format
- Submit the form
- You should be taken to a generic confirmation page

----

## Notes

- There is no validation (except enter a password) as validation will come for the backend to check if users current password is correct
- Not entirely happy with the code structure for having the text appear above the text input but could not find a cleaner way to do it (can revisit in the future when we refactor DisplayForm)


jennikate (approved):
LGTM

PR #182, opened 2023-02-07
NMSW-392 Update Cypress tests with new phone number pattern

jennikate:
- change error message wording to include errors that will appear for both fields
- change the phoneNumber input to now be a diallingCode and telephoneNumber

To Test
`npx cypress open --config-file cypress/config/dev.config.js --env MAIL_API_KEY=xxx`

replace xxx with the MAIL_API_KEY


Adele-B (approved):
👍

PR #183, opened 2023-02-07
NMSW-392 Require a number in phone number fields

jennikate:
# Ticket

NMSW-392

----

## AC

Fields with phone number validation currently allow you to enter only valid symbols and no numbers.
A number should be required

----

## To test

- log in
- go to `your details`
- go to `change your details`
- enter just symbols e.g. `()+`
- click submit
- > you should get an invalid format error
- enter letters
- click submit
- > you should see an error
- enter just numbers
- click submit
- > you should not see an error
- enter numbers and allowed symbols
- click submit
- > you should not see an error

----

## Notes

Added a test to check input display with type tel as well


Adele-B (approved):
LGTM

PR #186, opened 2023-02-09
NMSW-408 Cypress: Add body with registered email address to the DELETE user call

jennikate:
# Ticket

NMSW-408

----

## AC

The test user should be deleted after each registration test runs

----

## To test

Run Cypress registration tests

----

## Notes

Discovered endpoint now needs the Bearer token, and it's associated email address in the body of the request
Updated and tests should now pass


Adele-B (approved):
👍

PR #188, opened 2023-02-09
NMSW-401 Add case for expired token to the create voyage API call

Adele-B:
# Ticket

NMSW-401

----

## AC

Given my token has expired
When I click on report a voyage
Then I am directed to the sign in page

----

## To test

- Run the app
- Sign in
- Wait for token to expire (Devops can reduce token expiry time for testing)
- Click on report a voyage
- It should redirect you to the sign in page
- You should not see the token in session storage

----


jennikate (approved):
LGTM

PR #189, opened 2023-02-09
NMSW-268 Pass state in and out of sign in page on redirects

jennikate:
# Ticket

NMSW-268

----

## AC

In preparation for NMSW-268's file uploads for declarations
We need to be able to redirect a user into sign in and back to the voyage page, persisting their declaration id

----

## To test

- login
- start a declaration
- delete your token
- click submit button
- sign in
- > you should be returned to the general declaration upload page

----


Adele-B (approved):
LGTM

PR #190, opened 2023-02-09
NMSW-268 Pass state back out of Generic Message if exists

jennikate:
# Ticket

NMSW-268

----

## AC

Given I have passed state into the Generic Message page
When I click the 'next' link
That state should be passed on to the next page

----

## To test

Review and run unit tests as this does not have a UI enabled use case yet (coming in next story)

----


Adele-B (approved):
👍

PR #193, opened 2023-02-09
NMSW-387 Flash of cookie banner on page load

Adele-B:
# Ticket

NMSW-387

----

## AC

ISSUE
On initial page load, there is a flash of the cookie banner before the main page loads
 
EXPECTED BEHAVIOUR
Cookie banner should not flash
Cookie banner should only load if no cookie preference is found in browser cookies

----

## To test

- Run app
- In the code go to app
- Add `debugger` to the top of the App function
- Return to the app
- Click the arrow a couple of times
- > You should not see a flash of the cookie banner in the steps before the content loads
- Make sure the app works on page refresh AND when the network is set to slow 3G

----


jennikate (approved):
👍

PR #194, opened 2023-02-10
Remove unneeded linting

jennikate:
# Ticket

----

## AC

While doing a review of linting noticed we have some unneeded rules

----

## To test

- run tests
- run cypress tests
- go to `/create-account/your-details`
- > should redirect you to the Try again message
- go through create account process and click validate on your link
- > should should the Your details page

----

## Notes

- refactored a useEffect so it is not needed and the associated file can be removed : discovered when investigating `"react-hooks/exhaustive-deps": "off",` rule. There is more refactoring to be done so we can reset this back to the 'on' state
- removed some rules that are part of the AirBnB styles by default, we don't need to respecify them here
- We have two special cases where `"import/prefer-default-export": ["off"],` needs to apply, so have removed this rule as a general rule and overridden those two cases in their files (DownloadFile -- see comments on file it's causing us trouble with mocking for testing, and config file where we expect to be adding more config for this app)
- The only place we allow jsx in a .js file is index.js so removed `"react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],` and overridden in index.js
- `"react/jsx-no-useless-fragment": [2, { "allowExpressions": true }],` was in place for one file - ScrollToTopOnNewPage which takes {children} with no other jsx, so removed rule and we override on that file only

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -30,7 +30,6 @@
     "jsx-a11y/label-has-associated-control": ["error", { "assert": "either" } ],
     "max-len": ["error", { "code": 250, "ignoreStrings": true, "ignoreComments": true, "ignorePattern": "^path " }],
     "no-plusplus": ["error", { "allowForLoopAfterthoughts": true }],
-    "quotes": [ "error",  "single"],
     "react-hooks/exhaustive-deps": "off",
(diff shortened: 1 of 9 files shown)


Adele-B (approved):
LGTM

PR #196, opened 2023-02-16
NMSW-407 Add country and dialcode autocomplete inputs

jennikate:
# Ticket

NMSW-407

----

## AC

**Scenario One: phone dial code**

Given I am viewing the phone dial code field
When I have typed at least one number in the field
Then a list of possible matches is returned

Given I have typed at least one number in the phone dial code field
When there are no dial codes that match this number
Then I am shown no results found message

 

**Scenario Two: country**

Given I am viewing the country field
When I have typed at least 2 letters in the field
Then a list of possible matches is returned

Given I have typed at least two letters in the field
When there are no country names that contain this combination of letters
Then I am shown no results found message

----

## To test

- create an account
- go as far as the your details page

**Scenario One: dial code**

- type `+` and a number into the international dial code field
- > you should be shown options that match your number
- delete what you typed
- type a single number into the international dial code field
- > you should be shown options that match your number
- delete what you typed
- type a bunch of numbers that don't match a dial code
- > you should be shown no results found message
- delete what you typed
- type a number
- select an option from the suggestions
- > the combo box should hide and your number be showing as the value in the field
- refresh your page
- > your selection should persist

**Scenario Two: country**

- type two letters into the country field
- > you should be shown options that start with these two letters
- > you should be shown options that contain these two letters anywhere in the name
- > the results should remain sorted alphabetically based on the start of the country name
- type more letters until no countries match 
- > you should be shown no results found message
- delete what you typed
- type part of a valid country name
- select a country from the list
- > the combo box should hide and your country be showing as the value in the field
- refresh your page
- > your selection should persist

**Repeat on the edit your details page**

----

## Notes
does not support user typing '+' into the dialling code field - investigating for a separate PR


Adele-B (approved):
LGTM

PR #198, opened 2023-02-20
NMSW-426 Give webpack access to env variables

jennikate:
# Ticket

NMSW-426

----

## AC

Enable webpack access to env variables

----

## To test

Make sure you have `NMSW_DATA_API_BASE_URL` setup in your .env file

Add the following to the index.html file (this will console log the api env variable - and allows you to see it is retrieved through webpack and passed to index.html)

```
<script type="text/javascript">
    console.log('API', '<%= htmlWebpackPlugin.options.api %>')
  </script>
  ```

- `npm start`
- open app
- > in the console you should see the value of the api variable as the same value as your env file
- go to network tab
- click on `start now` button
- > in the network tab requestURL for the call, the api section of it should match the value in your env file (this tests that the env variable value is correctly passed to the config.js file still)

----

## Notes

This PR
- updates package.json and webpack to access variables from .env files
- updates the Dockerfile & nginx/run.sh to add the GA_TOKEN replace
- updates extract-env.sh and drone.yml to create a .env file and read the variables from there


Adele-B (commented):
Looks good from my side - webpack can access env variables and pass them into index.html
Handing over to @josh-samuel for review of DevOps related code and final approval! 😁

PR #202, opened 2023-02-21
NMSW-420 Allow '+' for dialCode autocomplete field

jennikate:
# Ticket

NMSW-420

----

## AC

Given a user types a `+` and a number into the international dialling code field
When we return the search results
They should ignore the `+` and still return all matching numbers

----

## To test

- go to edit your details
- type + into the dial code field
- type 61 into the dial code field
- > you should see the result for Australia

Repeat the above in the create account your details page

----


Adele-B (approved):
LGTM

PR #206, opened 2023-02-23
NMSW-435 Autocomplete should persist data on refresh & update cypress test to click selection

jennikate:
# Ticket

NMSW-435

----

## AC

Given I have selected an autocomplete item
When I refresh the page
The expanded details should persist

e.g.
Given I am on register your details step
When I have selected a country
And I refresh the page the country should remain the same
And I submit the form
And I enter a password
And I submit that form
Then the 3 digit country code should be passed to the API

----

## To test

- go to create account your details step
- enter/select an int dialling code and country
- refresh the page
- submit the form
- > it should not error on these fields
- enter passwords
- submit the form
- > the success page should be shown
- > you should see in the API payload we sent a telephone number in the format of (xx)xxxx and a country 3 letter alphacode

----


Adele-B (approved):
👍

PR #213, opened 2023-02-28
NMSW-438 Implement a fix for autocomplete bug

Adele-B:
# Ticket

NMSW-438

----

## AC

This fixes a bug where if on a page with an autocomplete component and the user enters their details but then deletes something in an autocomplete field, it will remove all the values of all the fields in sessionStorage - this causes all the fields to fail on submit even when it appears as if there are values in the inputs.

This relates to another bug where if a user creates an account and then goes back to the create-account/your-details page, when they enter more details and submit, all the fields will error. This could be related to the cypress tests erroring as we see a similar pattern in the failed tests 

Fixed the autocomplete but and also added a scenario to cover the user going back to the form and entering their details again but it will take them to the account already active page instead of the something went wrong page which would redirect to the password page over and over again.

----

## To test

- Run app
- Go to localhost:3000
- Create an account
- Get to the your details page
- Open devtools and go to the application tab to view sessionStorage
- Enter details into the name, company and dialling code fields
- Delete the value in the autocomplete field
- > The session storage should still show the values for the name and company fields but the dialling code values should be null
- Fill in the rest of the data
- Click continue
- > You should be taken to the password page and be able to successfully create an account
- Use the browser back button to go back to the /your-details page
- Fill in the details and password again
- Click continue
- > You should be taken to the Account already active page 
----

## Notes

- Clearing the state on the confirmation page does not work because when the user clicks the browser back button the email in the state returns


jennikate (approved):
LGTM

PR #217, opened 2023-02-28
NMSW-436 [pt4] Error mapper for turning API response into user facing error messages

jennikate:
# Ticket

NMSW-436

----

## AC

Translate field level errors received from API into user facing error messages

----

## To test

Unit tests should pass

----


Adele-B (approved):
LGTM

PR #218, opened 2023-02-28
NMSW-436 [pt5] Hande FAL1 General declaration missing mandatory fields inc Sorting

jennikate:
# Ticket

NMSW-436

----

## AC

When there are required fields with null values

----

## To test

- create a voyage
- upload FAL1 sheet from Jira ticket
- click submit
- > you should see the errors listed in the Jira ticket
- > the errors should be in the order of 'row' then 'column' (see image on Jira ticket for confirmation)

----


Adele-B (approved):
LGTM

PR #219, opened 2023-02-28
Fix template download path/name

jennikate:
# Ticket

----

## AC

Downloading FAL 1 from voyage report page was downloading an invalid version
We were using an old name, updated to current filename

----

## To test

Download FAL 1 template from voyage report General Declaration page

----


Adele-B (approved):
👍

PR #222, opened 2023-03-01
NMSW-441 Add error mapping for invalid characters in FAL 1 files

jennikate:
# Ticket

NMSW-441

----

## AC

When there are fields with invalid characters

-- | -- | -- | --
B3 | nameofship | contains character that is not alphanumeric | Enter the ship's name using English letters instead of special characters not recognised
F3 | callsign | contains characters that is not alphanumeric | Enter the call sign using only letters and numbers
B4 | signatory | contains character that is not alphanumeric | Enter the name of the master using English letters instead of special characters not recognised
B13 | cargo | alpha characters only | Enter the description using English letters instead of special characters not recognised

----

## To test

- upload a FAL 1 spreadsheet that has invalid characters in the fields listed above
- submit the file
- > you should see an error page with the above error messages
- > errors should be listed in row then column order (see image in Jira)

----


Adele-B (approved):
LGTM

PR #225, opened 2023-03-01
NMSW-443 Error mapping for arrivals to UK

jennikate:
# Ticket

NMSW-433

----

## AC

When the departurefromUK field is 'Arrival in the UK'
And related cells have invalid values

-- | -- | -- | --
B5 | arrivalport | If B2 = arrival then the first 2 letters must be GB | You selected arrival in the UK. Enter an arrival LOCODE that starts with GB
  |   | if B2 = arrival then arrival date must be AFTER departure date | You selected arrival in the UK. Enter an arrival date and time that is after the departure from the previous port
B6 | departureport | if B2 = arrival then first two letters should NOT be GB | You selected arrival in the UK. Enter a departure LOCODE for the last port of call and not a GB port
D7 | previousPort | if B2 = arrivals then value must match value of B6 | Enter a last port of call that matches the last port of call you gave in cell B6
F7 | nextPort | must NOT start with GB | Enter a LOCODE that does not start with GB for the next port of call

----

## To test

- upload a test file that has the errors shown above
- submit the file
- you should see the errors listed above

----

## Notes
date/time formatting is not yet tested


Adele-B:
👍

PR #226, opened 2023-03-01
NMSW-444 Error mapping for departures from the UK

jennikate:
# Ticket

NMSW-444

----

## AC

When the departurefromUK field is 'Departure from the UK'
And related fields are invalid

-- | -- | -- | --
B5 | arrivalport | If B2 = departures then the first two letters must NOT be GB | You selected departure from the UK. Enter an arrival LOCODE for the next port of call and not a GB port
B6 | departureport | if B2 = departure then first two letters should be GB | You selected departure from the UK. Enter an departure LOCODE that starts with GB
  |   | if B2 = departure departure date must not be more than now - 1 day | Enter a departure date from the UK that is not more than 24 hours in the future
  |   | if B2 = departures then departure must not be before date of arrival | You selected departure from the UK. Enter a departure date and time that is before the arrival at the next port
D7 | previousPort | must NOT start with GB | Enter a LOCODE that does not start with GB for the last port of call
F7 | nextPort | if B2 = departures then value must match value of B5 | Enter a next port of call that matches the next port of call you gave in cell B5

----

## To test

- upload test files that contain the errors above
- submit the file
- you should see the errors listed above

----

## Notes
date/time formatting not tested yet


Adele-B (approved):
LGTM

PR #227, opened 2023-03-01
Update docs with error mapping

jennikate:
Update the docs to include error mapping option for file uploads
& fix some formatting


Adele-B (approved):
👍

PR #235, opened 2023-03-03
NMSW-447 Show helpful error on 500 when uploading FAL forms as it's likely to be due to badly formatted date/time

jennikate:
# Ticket

NMSW-447

----

## AC

For MVP, until we have the additional validation in the BackEnd for fields (coming in NMSW-350) a field that requires date/time format, that is in an invalid format, will return a 500 error.

We want to show the user an error message that indicates this is the likely cause of the 500 error so they can try correcting that.

----

## To test

- start a voyage
- for the FAL1 general declaration page upload a file with a bad date/time format
- click submit
- > you should get the message page with a message about checking date time formats
- > the page should have a re-upload button
- click the re-upload button
- > you should be returned to the FAL 1 upload page

----


Adele-B (approved):
LGTM

PR #237, opened 2023-03-03
NMSW-451 Fix the styling for displaying the voyage reports 

Adele-B:
# Ticket

NMSW-451

----

## AC

We have noticed lots of styling inconsistencies as they layout did not match the GDS grid system of building pages. Therefore, it will need to be rebuilt so it can look and function as intended.

----

## To test

- Run app
- Go to localhost:3000
- Sign in
- The voyages layout displayed /your-voyages page should be a better match to the prototype

----


jennikate (approved):
LGTM

PR #238, opened 2023-03-06
NMSW-388 Handle error responses on GET declaration

jennikate:
# Ticket

NMSW-388

----

## AC

Given FE has made a GET request to the /declaration/<declarationId>/attachments endpoint
When the response is 401
Then the user is taken to the sign in page
When the user signs in
Then they are returned to the task details page

Given FE has made a GET request to the /declaration/<declarationId>/attachments endpoint
When the response is 422
Then the user is taken to the sign in page
When the user signs in
Then they are returned to the task details page

Given FE has made a GET request to the /declaration/<declarationId>/attachments endpoint
When the response is 500
Then the user is taken to the message page
When the user clicks on the continue link
Then they are taken to the Your Voyages page

----

## To test

- create a voyage
- upload a general declaration file & submit it
- delete your auth token
- click button to go to next page
- sign in
- > you should be taken to task details page
- > you should see the ship name and voyage type filled in

----


Adele-B (approved):
LGTM

PR #240, opened 2023-03-06
NMSW-448 Disable input and submit on submit click for file upload until process complete

jennikate:
# Ticket

NMSW-448

----

## AC

Given user has chosen a file
When they click check for errors button
Then the input field and the button should be disabled

Given the input field and the button are disabled
When there are errors on the page (e.g. invalid file type)
Then the input field and the button should no longer be disabled

----

## To test

- put your browser on slow 3G so you can see the changes more easily
- create a report
- choose an image (or other invalid type) file
- click check for errors
- 
- > the check for errors button should be disabled and you cannot click it
- > the input / choose file field should be disabled and you cannot click it
- wait for the error to load
- > the check for errors button should be active and clickable
- > the input / choose file field should be active and clickable

----

## Notes
- refactored error handler to remove some repetition
- added comments about the temp handle 500 error solution in place until BE can return a better error for date/time format issues
- added a fallback for if we get an unexpected success status


Adele-B (approved):
LGTM

PR #243, opened 2023-03-06
Test nmsw 438

JayanthiiB:
- Test to verify deleting the previous value in auto-complete fields not losing other field values
- Added test to verify Account already active page when navigating back again from the password page

## To test
```
npm run cypress:open:dev -- --env MAIL_API_KEY=<API_KEY>
```
and select registration.feature


jennikate (approved):
LGTM

PR #250, opened 2023-03-10
NMSW-469 Add missing code to remove auth token when forcing a sign out

Adele-B:
# Ticket

NMSW-469

----

## AC

We have noticed that in some places we are missing a line of code (`Auth.removeToken()`) which removes the auth token from session storage when we force a sign out when the BE sends expired/missing segments token message

Adding this code where it is missing (File upload pages, task list page and your voyages page) so that if a user is signed out they are not able to access/see any protected routes

----

## To test

- Run app
- Go to localhost:3000
- Sign in
- Change token value in session storage to 123
- Click on Report a voyage
- > You should be taken to the sign in page
- > The token should be removed from session storage 
- > AND you should not see the nav or be able to access protected routes e.g. /your-voyages
- Sign back in
- Click on report a voyage
- Upload a file
- Change auth token value to 123
- Click check for errors
- > You should be taken to the sign in page
- > The token should be removed from session storage 
- > AND you should not see the nav or be able to access protected routes e.g. /your-voyages
- Finally Sign back in
- Click on report a voyage
- Upload a file with good data
- Click Check for errors 
- On the upload success page change auth token value to 123
- Click save and continue
- > You should be taken to the sign in page
- > The token should be removed from session storage 
- > AND you should not see the nav or be able to access protected routes e.g. /your-voyages
----


jennikate (approved):
👍

PR #259, opened 2023-03-14
NMSW-463 Multi file upload POST & status setting functions

jennikate:
# Ticket

NMSW-463

----

## AC

Given I have selected one or more files
When I click upload files
Then the first files status changes to in progress
When the file is finished uploading
Then it's status changes to a complete status
And the next file changes to in progress
When that file is finished uploading
Then it's status changes to a complete status
And this is repeated until all selected files are complete

Given a file finishes uploading
When the upload was a success
Then it's status is shown as complete-success

Given a file finishes uploading
when the upload failed
Then it's statue is shown as complete-err

----

## To test

- go to supporting docs page
- select two or more FAL 1 files (at least one should be good, one error)
- click upload files
- > you should see the first file change status to in progress
- > and then to a complete status
- > and the second file change to in progress
- > and then to a complete status
- > this should repeat until all files are done

----

## Notes

- this is a test endpoint only, using FAL1 endpoint to test concept, so only FAL1 files can be added during this test phase

- the status names will be changed in later tickets to something user friendly


Adele-B (approved):
LGTM

PR #262, opened 2023-03-16
NMSW-463 UI for file upload statuses & handle missing/invalid auth token

jennikate:
# Ticket

NMSW-463

----

## AC

Given I have selected two or more files
When I click 'upload'
Then the first file should start uploading 
And its pending tag be removed
And a loading spinner show for it

When a file receives a Success response
Then that file's loading spinner should be removed
And it should show a green tick icon
And it should show 'has been uploaded' next to the file name

When this is not the last file
Then the next file should start uploading
And its pending tag be removed
And a loading spinner show for it

When a file receives an Error response with a message
Then that file's loading spinner should be removed
And it should show a red cross icon
And the error message should show next to the file name

When a file receives an Error response without a parsable message
Then that file's loading spinner should be removed
And it should show a red cross icon
And the generic error message of 'WORDING TBC' should show next to the file name

When this IS the last file
Then all the files should have a status that is success or error

----

## To test

- create a voyage
- go to supporting documents page
- select at least
- - one good FAL 1 form (will return success)
- - one FAL 1 form with invalid dates (will return a 500)
- - one file that is not xlsx (will return 400 invalid file type)
- > you should all files with a pending tag

click upload
> you should see a loading spinner next to the first file 
> when it completes you should see the relevant status message
> and the next file should have a loading spinner
> when that completes you should see the relevant status message
> and the next file should have a loading spinner
> this repeats until all files are done

Scenario: no auth token/auth token invalid (Automated testing)

- create a voyage
- go to supporting documents page
- add at least one file
- remove the Auth token
- click upload
- > you should be redirected to sign in page
- sign in
- > you should be returned to the file upload page (see notes below)

Scenario: Auth token lost part way through upload (QAT test)

- create a voyage
- go to supporting documents page
- set your network to slow 3g so you have time
- add a few files
- click upload
- once the first file changes to a loading spinner, delete your auth token
- > the first file should return a success/error message and this be displayed
- > the second file should start uploading
- > before it shows a status you should be redirected to the sign in page

----

## Notes

- we know the auth token is invalid when we get a response from the POST attempt
- we can't persist any files that were in a 'pending' state in the browser as there are browser security in place to prevent this, so the user will have to select all their files again
- We WOULD normally show the user the files that were successfully uploaded, but the GET endpoint for getting files doesn't exist yet so we can't get a list of successfully uploaded files


Adele-B (approved):
LGTM

PR #263, opened 2023-03-16
NMSW-463 Allow max 7 files to be added to supporting docs

jennikate:
# Ticket

NMSW-463

----

## AC

Given I am selecting supporting documents for upload
When I select more than 7
Then I am shown an error

Given I have selected at least 1 supporting document (e.g. select 2)
And it is shown in the list of selected files
When I select more and the total is now more than 7 (e.g. select 6, total is 8)
Then I am shown an error
And I can see the originally selected files in the list
And I cannot see the new files I tried to add in the list

----

## To test

- crete a voyage
- go to supporting docs
- select 8 or more files
- > you should see an error
- select 2 files
- > you should see those two files appear in the file list
- select another 6 files
- > you should see an error
- > you should see the two original files in the file list
- > you should not see the new files you tried to add

----

## Notes

NOTE: once we build the final POST and GET endpoints then we will extend this feature to include previously uploaded files, but that is not in place for this ticket so will not be tested here


Adele-B (approved):
LGTM

PR #265, opened 2023-03-16
NMSW-499 Add wording for if max files exceeded

jennikate:
# Ticket

NMSW-499

----

## AC

Given I have previously uploaded or currently selected 0 files
When I select 8 or more files
Then I am shown the error "You've selected too many files. You can only add 7."

Given I have previously uploaded or currently selected 1 or more files
When I select more files than allowed
Then I am shown the error "You've selected too many files. You can only add x more files"

e.g.
Previously uploaded 1 file
Selected 7 files
Error: You've selected too many files. You can only add 6 more files.

----

## To test

- create a voyage and go to supporting documents
- try to add 8 files
- > you should see the error "You've selected too many files. You can only add 7."
- add one file
- now try to add 8 more files
- > you should see the error "You've selected too many files. You can only add 6 more files"

----


Adele-B (approved):
LGTM

PR #268, opened 2023-03-17
Add redirect to resend verification email if unexpected error

jennikate:
# Ticket

NMSW-472

----

## AC
Bug details
**To recreate**

- register a user, but do not verify them
- go to sign in
- attempt to sign in with that email address

**Expectation**
you get the 'verification email resent' message

**Actual**
you get a something went wrong error

----

## To test

- register an email address but DO NOT verify it
- go to sign in
- attempt to sign in with that email address
- > you should see the 'verification email sent' page
- > you should receive a verification email

----

## Notes
The API response for this scenario is currently 500, there is a ticket open to improve the response received but until that is done this code will cover the scenario.


Adele-B (approved):
LGTM

PR #269, opened 2023-03-20
NMSW-470 p1 Refactor use of declarationId in General Declaration FAL 1 upload

jennikate:
# Ticket

NMSW-470

----

## AC

Refactor declarationId to be on URL params so we can later redirect users out to sign in and back to voyage without needing to persist state and session when they become unAuthorised due to token expiry/invalidation

----

## To test

To test refactored elements

- create a new voyage
- > on general declaration page in the url you should see the declaration id
- upload a bad file
- click check for errors
- > you should see error page (note, it won't have the declaration id in the url)
- click the re-upload file button
- > you should be returned to the general declaration page, with the declaration id in the url
- upload a good file
- click check for errors
- > you should be taken to the success page
- click save and continue
- > you should be taken to the task details page
- > the ship name and voyage type should match what you uploaded

----

## Notes

NOTE: 
- task details page & check your answers pages not yet updated so you cannot click the 'general declaration link in them'
- the url that includes the declaration id is a protected route so if user is not signed in and attempts to go directly to a url with a declaration id they will be automatically redirected to the landing page


Adele-B (approved):
LGTM

PR #274, opened 2023-03-20
NMSW-470 p5 Refactor declarationId - Delete draft

jennikate:
# Ticket

NMSW-470

----

## AC

see https://github.com/UKHomeOffice/nmsw-ui/pull/269

----

## To test

- create a voyage report, upload a gen dec file, and get to task details
- click the delete button
- > you should be taken to the do you want to delete voyage page with the declarationId in the url
- select no and save
- > you should be returned to your task details page
- click the delete button
- select yes and save
- > you should be taken to the your voyage page
- > TEMP: you should see a console log confirming delete of ship name (this will be replaced with a confirmation banner once we have a delete endpoint and can delete)

----

## Notes

The CYA page will not work yet as it has not been refactored


Adele-B (approved):
LGTM

PR #278, opened 2023-03-21
NMSW-501 Add findBy for 'There is a problem'

jennikate:
# Ticket

NMSW-501

----

## AC

ISSUE
unit test fails due to the assertion that 'There is a problem' should be in the document

We identified this is happening as the assertion is running before the error component of the page has finished loading

SOLUTION
We can add a findBy for that text, which forces a wait for the text to be available before it errors

This reduces the chance of the unit test failing just because it hasn't waited long enough for the component to load

----

## To test

Run unit tests

----


Adele-B (approved):
LGTM

PR #282, opened 2023-03-23
NMSW-394 Update error messages

Adele-B:
# Ticket

NMSW-394

----

## AC

UCD has reviewed all error messages and need to update app & tests to match. We're switching to 'harsher' validation error messages. So 'Enter a password that does not contain spaces' is changing to 'Password must not contain spaces'. This is in order to align more closely with COP and E@B error messages.

----

## To test

- Open spreadsheet attached to the ticket and scroll down to the current NMSW error messages
- Run app
- Compare the error messages marked as 'incorrect' on the spread sheet to the error messages on the app
- > The error messages in the app should match the new error messages in the table on the spreadsheet
- Check FAL 5 error messages for invalid and max characters (see note below)

----

## Notes

- **I have had to make changes to the FAL 5 error mapping due to a BE refactor - will need to check FAL 5 errors for invalid characters and max characters**

### DIFF
--- a/docs/form_creator.md
+++ b/docs/form_creator.md
@@ -318,7 +318,7 @@ It should NOT be used for the other types as it is redundant for them.
   validation: [
     {
       type: VALIDATE_REQUIRED,
-      message: 'Select is your company a shipping agent',
+      message: 'Select yes if your company is a shipping agent',
     },
   ],
 },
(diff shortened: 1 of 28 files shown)


jennikate (approved):
LGTM

PR #284, opened 2023-03-23
NMSW-502 Allow 8 file to be added in supporting documents

Adele-B:
# Ticket

NMSW-502

----

## AC

We send FAL5&6 as one file, and FAL1 as one files - so can allow user to upload up to 8 more files.

----

## To test

- Run app
- Sign in
- Click Report a voyage
- Upload a good FAL 1
- Get the the task list page
- Click on supporting documents 
- Select 8 files
- > All 8 files should show in the list
- try to add a 9th file
- You should get max file error
- Refresh the page to clear the errors
- Try to add 9 files
- > You should get max file error
- Refresh the page to clear the errors
- Add a file
- Then try to add another 8 files
- You should get a max file error

----


jennikate (approved):
LGTM

PR #292, opened 2023-03-27
NMSW-470 Redirect protected pages to sign in when no auth token

jennikate:
# Ticket

NMSW-470

----

## AC

Given a user tries to access a protected url
When the user is not signed in
Then they will be directed to the sign in page
When they sign in successfully
Then they will be redirected back to the protected url they tried to access

Given a request is made to a declaration endpoint
When the response indicates the declaration does not belong to the signed in user (404 declaration does not exist)
Then a message of 'declaration does not exist' is shown

----

## To test

This impacts all logged in pages, and the entire report voyage journey.
The key scenarios are listed in the Jira ticket.

To check the code changed in this ticket specifically you can

- start a report
- add a bad general declaration file
- click check for errors
- > you should see the error page
- delete your auth token
- click re-upload file
- sign in
- > you should be returned to the upload general declaration page
- add a good file
- click check for errors
- > you should be taken to the success page
- delete your auth token
- click save and continue
- > you should be taken to sign in
- sign in
- > you should be taken to task details page
- delete your token
- click on check your answers
- > you should be taken to sign in
- sign in
- > you should be taken to check your answers

----

### DIFF
--- a/src/pages/Voyage/VoyageDeleteDraftCheck.jsx
+++ b/src/pages/Voyage/VoyageDeleteDraftCheck.jsx
@@ -51,7 +51,6 @@ const VoyageDeleteDraftCheck = () => {
         },
       ],
     },
-
   ];
 
   const handleSubmit = (formData) => {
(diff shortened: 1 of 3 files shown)


Adele-B (approved):
LGTM

PR #299, opened 2023-03-29
NMSW-522 Update FAL 5 erroring for blank files and field level errors

Adele-B:
# Ticket

NMSW-522

----

## AC

Given I try to upload a blank FAL 5 file
When I click check for errors
Then I am shown a 'Template is empty' error in the error summary

Due to BE changes and after speaking with UCD - blank files should now return a 'Template is empty' error that will show in the error summary on the Crew Details upload page

Also, the wording we get from the BE for certain errors has been changed so we will also need to re-map these errors on the FE

- Blank field errors
- Max character errors
- Invalid character messages

----

## To test

- Run app
- Create a voyage
- Get to Crew details FAL 5 upload
- Upload a blank file
- > You should get a 'Template is empty' error in the summary
- Upload a file with invalid characters
- > You should be taken to the error page with invalid character errors
- Click re-upload file
- Upload a file with max characters
- > You should be taken to the error page with max character errors
- Upload a file with blank fields in some of the required fields
- > You should get field is required errors

----

### DIFF
--- a/src/components/FileUploadForm.jsx
+++ b/src/components/FileUploadForm.jsx
@@ -6,6 +6,7 @@ import { GENERAL_DECLARATION_TEMPLATE_NAME, MAX_FILE_SIZE, MAX_FILE_SIZE_DISPLAY
 import {
   FILE_MISSING,
   FILE_TYPE_INVALID_PREFIX,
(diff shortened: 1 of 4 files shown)


jennikate (approved):
lgtm

PR #304, opened 2023-04-03
NMSW-529

jennikate:
# Ticket

NMSW-529

----

## AC

- user pastes a URL for a file upload successful page
- user is not signed in
- user signs in

**expected:**
we redirect to file upload success page

**actual:**
we redirected to something went wrong

This was because we required filename to be in state for file upload success page, and in this above scenario it is not.

----

## To test

- create a voyage up to a file upload success page
- copy the url
- sign out
- paste the url
- sign in
- you should be returned to the file upload success page

----

## Notes

If you paste this url and sign in as a user who does not own that declaration ID you will be shown the file upload success page. However when you click 'save and continue' you will get an error. Ticket 530


Adele-B (approved):
LGTM

PR #307, opened 2023-04-05
NMSW-531 Display draft voyages

Adele-B:
# Ticket

NMSW-531

----

## AC

GIVEN a voyage has been started, but has not been submitted
WHEN I view the voyage list
THEN the voyage is displayed as "DRAFT"

----

## To test

- Run app
- Sign in
- > If you have reports attached to your account you should see a list of draft reports with a continue button next to each
- > You should not see reports that do not have a FAL 1 attached (reports in the list without ship name, voyage type or a date)
- Click on continue
- > You should be taken to the task list page for that voyage (only for drafts)
- Go back to the your voyages page
- Delete your token and refresh
- > You should be taken to the sign in page

----

## Notes

- We have removed the filtering for MVP - new design has been okayed by UCD team
- We only get draft reports from the BE so only voyages with the status of draft will be seen
- Drafts without an attached FAL 1 still exist but are filtered out - this code will be changed once we can delete voyages
- The line at the bottom of each report on the prototype, for example, 'Submitted: {DATE} by {PERSON}' has been taken out, as this adds to the complexity when dealing with draft and cancelled voyages needing different dates from the BE for sorting so has been removed for MVP

### DIFF
--- a/src/pages/NavPages/YourVoyages.jsx
+++ b/src/pages/NavPages/YourVoyages.jsx
@@ -229,14 +229,14 @@ const YourVoyages = () => {
                     <div className="govuk-grid-row">
 
                       <div className="govuk-grid-column-one-quarter reported-voyages__columns reported-voyages__text">
-                        <p className="govuk-body-s govuk-!-font-weight-bold">{voyage.shipName}</p>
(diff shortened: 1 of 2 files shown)


jennikate (approved):
LGTM

PR #311, opened 2023-04-06
NMSW-533 Update 'do you have passengers' question to PATCH response to endpoint

Adele-B:
# Ticket

NMSW-533

----

## AC

Given a user has selected yes or no to the 'do you have passengers' question
When they click save
Then we PATCH that to the endpoint

----

## To test

- Run app
- Sign in
- create a voyage
- go to passengers page
- select YES
- click save
- > check the network tab, you should see a successful PATCH to the endpoint with the value of TRUE for passengers
- go back to the passengers page
- select NO
- click save
- > check the network tab, you should see a successful PATCH to the endpoint with the value of FALSE for passengers

----


jennikate (approved):
LGTM

PR #317, opened 2023-04-12
NMSW-204 Request forgot password reset link

jennikate:
# Ticket

NMSW-204

----

## AC

Given I have entered a registered and validated email address
When I click submit
Then I should be shown the confirmation page
~~And I should receive a password reset email~~

Given I click 'not received an email' on the confirmation page
Then I should be taken to the request a new link page

Given I click request a new link on the confirmation page
When the stored email address is a registered and validated one
Then I should be shown the confirmation page
~~And I should receive a password reset email~~

Given I enter an email address that is not registered and validated on the forgot password page
When I click submit
Then I should be shown the confirmation page
But no email should be sent

----

## To test

**NOTE: to get to the request a new email page for testing UI purposes change the '401' error handler to '400'**

SCENARIO 1: registered and validated email : can reset their password

- go to /forgotten-password
- enter a registered and validated email address
- > ~~you should receive a password reset email (delete this for now)~~
- click on not received an email
- > you should be taken to the request a new email page

SCENARIO 2: cannot reset password as email not registered and validated

- go to /forgotten-password
- click submit
- > you should receive an error on the page
- enter a badly formatted email
- click submit
- > you should receive an error on the page
- enter a correctly formatted email that is not registered and validated
- check the network tab
- > you should see a 401 error, which indicates there was no email sent

----

## Notes
- we will not receive a 204 success response at the moment as the required template ID for the email on dev/staging isn't setup correctly
- the 'request a new link' page hasn't been built out yet, that comes in the next PR
- added a fallback for the 'message' page for if a message is received from the API that is not in a string format as we got one and the page broke

### DIFF
--- a/src/AppRouter.jsx
+++ b/src/AppRouter.jsx
@@ -28,6 +28,7 @@ import {
   REGISTER_DETAILS_URL,
   REGISTER_PASSWORD_URL,
   REQUEST_PASSWORD_RESET_URL,
+  REQUEST_PASSWORD_RESET_CONFIRMATION_URL,
   SIGN_IN_URL,
   TEMPLATE_PAGE_URL,
   VOYAGE_CHECK_YOUR_ANSWERS,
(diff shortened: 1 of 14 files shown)


Adele-B (approved):
LGTM

PR #313, opened 2023-04-11
NMSW-154 Display voyage details on the page inc file downloads

Adele-B:
# Ticket

NMSW-154

----

## AC

GIVEN I have a link to an uploaded crew file
WHEN I click on the file name
THEN the file will be downloaded

GIVEN I have a link to an uploaded passenger file
WHEN I click on the file name
THEN the file will be downloaded

----

## To test

- Run app
- Sign in
- Complete General declaration FAL 1
- Complete Crew details (FAL 5)
- Click on Any passenger details
- Select no to passengers
- The CYA link should be enabled
- Click on the CYA link
- Scroll down to the file uploads sections
- > You should see a link to the FAL 5 you uploaded and have the ability to download it
- > It should be the same name that you uploaded
- > The passenger section should state 'No passenger details provided'
- > The supporting documents section should state 'No supporting documents provided'
- Click on change next to Passenger details
- Select yes
- Go back to CYA (do NOT upload a file)
- On the CYA page click submit
- > You should get an error of 'Passenger details (FAL 6) upload is required for ships carrying passengers' in the error summary
- Click on change next to Passenger details
- Select yes
- Upload a good FAL 6
- Return to the CYA page
- > You should see a link to the FAL 6 you uploaded and have the ability to download it
- > It should be the same name that you uploaded

----

PR #320, opened 2023-04-13
NMSW-554 Update FE code to accommodate BE changes to how we receive FAL data

Adele-B:
# Ticket

NMSW-554

----

## AC

We now receive FAL 5 and 6 data as a list of objects which means that the code for NMSW-424 and NMSW-154) is no longer working as expected.

FE code that needs updating:
- Tasklist page: Make sure that the statuses are updating as expected for FAL 5 and 6 uploads
- CYA page: Make sure the page loads and displays the filename and has a link with the ability to download the file as expected

----

## To test

- Follow test steps for https://github.com/UKHomeOffice/nmsw-ui/pull/312
- Follow test steps for https://github.com/UKHomeOffice/nmsw-ui/pull/313

----


jennikate (approved):
LGTM

PR #324, opened 2023-04-17
NMSW-244 Submit a declaration

jennikate:
# Ticket

NMSW-244

----

## AC

Given I have completed all required inputs
And I am on the Check Your Answers page
And my declaration status is draft
When I click 'submit'
Then a PATCH request is sent to /declarations with a status of PreSubmitted'

> from here the backend takes over with the submission to CBP and the sending of emails to the user/support as required

----

## To test

- create a declaration
- go to the CYA page
- click submit
- > you should see the confirmation page
- > you should see in network tab a PATCH request for that declaration with a PreSubmitted status

----

## Notes
Added the standard handling of 422/401/500 responses from the PATCH
We only have 'draft' status declarations at the moment - different buttons/functions will be added for submitted status (cancel / handleCancel), and no button/function if status is cancelled

### DIFF
--- a/src/components/ConfirmationMessage.jsx
+++ b/src/components/ConfirmationMessage.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { SERVICE_NAME } from '../constants/AppConstants';
 
 const ConfirmationMessage = ({
+  children,
   confirmationMessage,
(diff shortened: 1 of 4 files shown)


Adele-B (approved):
LGTM

PR #327, opened 2023-04-18
NMSW-243 Display uploaded supporting documents

Adele-B:
# Ticket

NMSW-243

----

## AC

GIVEN I have previously uploaded supporting docs
WHEN I return to the supporting docs page
THEN I am shown a list of previously uploaded docs

----

## To test

**Scenario 1: Success case**
- Run app
- Sign in
- Create a declaration
- Complete General declaration FAL 1 section
- On the task list page click supporting documents link
- Add a couple of valid files (csv, doc, docm, docx, rtf, txt, xls, xlsm, xlsx, xltm, xltx, xlw or xml AND a file size less than 1MB)
- Click upload
- > They should successfully upload
- Click the back button
- Return to the supporting documents upload page
- > You should see the files you previously uploaded

**Scenario 2: Max file errors**
- Continue from the last part of success case (already have uploaded files)
- Select too many files (number depends on how many you uploaded in previous step)
- > You should see an error stating the remaining files
- Add 2 acceptable files
- Click upload (do NOT refresh - the 2 uploaded files will be in a 'limbo' state)
- Try to select too many files
- > You should see an error stating the remaining files (it should have included the 2 limbo files in the total)

**Scenario 3: File already exists**
- Continue from scenario 2
- Select one of the same files from scenario 1 
- > You should get a file already exists error
- Select one of the same files from scenario 2 (a file in limbo state)
- > You should get a file already exists error

**Scenario 4: Other features**
- Continue from scenario 3
- Select a file that is greater than 1MB
- > You should see an error regarding file size
------------------------------------------
- Select 1 file
- Click upload
- Select 1 more file
- Click the upload button again
- > The first file/any file in the 'limbo' state should not upload a second time, but the second file should upload as normal
----


jennikate (approved):
LGTM

PR #331, opened 2023-04-19
NMSW-514 Delete uploaded supporting document

Adele-B:
# Ticket

NMSW-514

----

## AC

Given a user has uploaded a file
When they click delete
Then the file is removed

----

## To test

**To test deleting previously uploaded file**
- Run app
- Sign in
- Create a declaration
- Complete General declaration FAL 1 section
- On the task list page click supporting documents link
- Add a couple of valid files (csv, doc, docm, docx, rtf, txt, xls, xlsm, xlsx, xltm, xltx, xlw or xml AND a file size less than 1MB)
- Click upload
- Refresh the page
- Click delete
- > You should see a successful delete request in the network tab

**To test deleting file in 'limbo' state**
- Add another valid file
- Click upload
- Do NOT refresh the page
- Click delete
- > You should see a successful delete request in the network tab

----

## Notes

- Currently the file will not disappear from the list because the BE the file gets deleted from S3 but does not get deleted from the database


jennikate (approved):
There's a slightly odd behaviour after the DELETE, the GET result ends up showing the files that were last uploaded as duplicates in the UI.

It shouldn't happen - it's hard to test what's causing that at the moment due to the DELETE not actually deleting a file (it deletes from S3 but not the database atm)

Code LGTM so approving PR and we'll fix forwards if we find issues once the backend deletes files fully.

PR #332, opened 2023-04-20
Version 0.13.0

Adele-B:
Merge version 0.12.0 to main branch


jennikate (approved):
Some Cypress tests failing but will be fixed as part of next release branch.
Signing off with this as we're still pre-live

PR #335, opened 2023-04-20
Password reset scenario

jennikate:
# Ticket

NMSW-204

----

## AC

Given a user has registered but not verified their email
When they submit the forgotten password form
Then they are shown an info page asking them to verify their email

----

## To test

- register an email address
- go to sign in
- go to forgot password
- try to send a forgot password link
- > you should see the 'you need to verify page'

----

## Notes
Scenario was missed from original PR


Adele-B (approved):
LGTM

PR #336, opened 2023-04-20
Replace sign in unverified flow

jennikate:
# Ticket

NMSW-204

----

## AC

Given a user has registered but not verified their email
When they submit the sign in form
Then they are shown an info page asking them to verify their email

_Note: we had two other implementations of handling this scenario which are removed as part of this PR as once this new UX was designed for the forgotten password flow it was decided to use the same UX for sign in_

----

## To test

- register an email address
- go to sign in
- try to sign in with that email and any password
- > you should see the 'you need to verify' page

----


Adele-B (approved):
LGTM

PR #339, opened 2023-04-21
NMSW-538 Add status tag component and refactor to use it

jennikate:
# Ticket

----

## AC

We're using status tags across a few files now so add a status tag component and refactor existing places to use it

----

## To test

- view Your voyages
- > draft tags should appear as before
- create a declaration
- add a FAL1
- > task details page should show the completed, required, optional, and cannot start tags as before
- add a FAL5 and select no for passengers
- > CYA should now show a not started tag as before

----

## Notes
Not all statuses are used on a page yet so some are currently untestable in the UI but I've added them here in preparation for future PRs

### DIFF
--- a/src/components/StatusTag.jsx
+++ b/src/components/StatusTag.jsx
@@ -1,0 +1,76 @@
+import PropTypes from 'prop-types';
+import {
+  DECLARATION_STATUS_CANCELLED,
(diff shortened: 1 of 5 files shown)


Adele-B (approved):
LGTM

PR #342, opened 2023-04-21
Test NMSW-557 Show uploaded supporting documents on CYA page (with ab…

JayanthiiB:
Test covered for NMSW-557
Show uploaded supporting documents on CYA page with the ability to download
----

## To test
```
npm run cypress:open:dev -- --env MAIL_API_KEY=<API_KEY>
```
Select check-your-answers-page.feature


jennikate (approved):
LGTM - there is a problem with the FAL1 sheet causing an error unexpectedly. BE team are looking into it but the logic in this test is fine so approving merge

PR #346, opened 2023-04-24
NMSW-514 Delete supporting docs pt 2

Adele-B:
# Ticket

NMSW-514 (pt 2 now that BE is working)

----

## AC

A bug was found where if you upload a file, then delete that file, you see two instances of that file. This was because the files in 'limbo' state (uploaded but were not retrieved from the initial GET call) are still a part of the filesAddedForUpload array but on calling the getDeclarationData again, they also become a part of the supportingDocumentsList so they appear twice. This code does this by filtering files with a pending tag so that these do not disappear is you delete an uploaded file in it's 'limbo' state

----

## To test

Scenario 1: Deleting previously uploaded file 
- Run app
- Sign in
- Create a voyage declaration
- Get to supporting docs section
- Upload a couple of files
- Refresh the page
- Delete one of the files
- > It should be removed from the list and you should only see one file

Scenario 2: Deleting an uploaded file in 'limbo' state
- Add a couple more files
- Upload them (should be in the 'limbo' state)
- Delete one
- > It should disappear from the list
- > There should not be a duplicate file

Scenario 3: Deleting an uploaded file in 'limbo' when there is a file in 'pending' state
- Add a another file
- Upload it
- Add another file (should be in 'pending' state)
- Delete the first file
- You should still have a file in 'pending'

----

### DIFF
--- a/src/components/MultiFileUploadForm.jsx
+++ b/src/components/MultiFileUploadForm.jsx
@@ -83,10 +83,23 @@ const MultiFileUploadForm = ({
   const [maxFilesError, setMaxFilesError] = useState();
   const [supportingDocumentsList, setSupportingDocumentsList] = useState([]);
 
+  const getPendingFiles = () => {
+    const pendingFiles = filesAddedForUpload.reduce((results, fileToCheck) => {
+      if (fileToCheck.status === FILE_STATUS_PENDING) {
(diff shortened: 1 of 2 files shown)


jennikate (approved):
LGTM

PR #348, opened 2023-04-24
NMSW-245 On GET drafts delete any without general declarations

Adele-B:
# Ticket

NMSW-245

----

## AC

If a voyage is in draft status and does not have general declaration details then we send a delete request for it 

----

## To test

- Run app
- Sign in
- Go to Your voyages
- You should only see declarations with General declarations attached
- Open network tab
- Create a declaration (do not upload general declaration)
- Go back to your voyages
- You should see a successful delete request for the declaration you just created

----

## Notes

- I opened this as a draft because the delete request was running twice and one was failing as the declaration no longer existed - this has now been resolved by jennikate so I have also reviewed her changes


jennikate (approved):
👍

PR #350, opened 2023-04-25
NMSW-584

jennikate:
# Ticket

----

## AC

Given I have uploaded files previously
When I click delete next to one
The button should be disabled so I cannot accidentally click it again

----

## To test

- create a declaration
- go to supporting docs
- add some files
- upload those files
- make sure you're on slow 3G
- click delete next to one file
- > the button should be disabled while the delete processes

----


Adele-B (approved):
LGTM

PR #351, opened 2023-04-25
NMSW-552 Remap errors as wording has changed from API

jennikate:
# Ticket

----

## AC

Error wording was updated in API as part of the overhaul of error messaging
We still need to catch this particular error in front end as we display it as a form error, not the error list page error

----

## To test

You will need
- a FAL 5 where two or more rows have the same document number and country
- a FAL 6 where two or more rows have the same document number and country
- a FAL 5 which has all unique rows
- a FAL 6 which has all unique rows within the file, however at least one of the rows should have the same document number and country as a row on the unique FAL 5 document

- create a declaration
- upload a FAL 5 where two or more rows have the same document number and country
- > you should see the error `Details listed on this file are not allowed, because they're the same as details you've already uploaded. Check the details in your file and try uploading again.` above the form

- upload the FAL 5 that has unique rows
- go to passenger and select yes
- upload the FAL 6 that has unique rows other than one that is the same as FAL 5
- > you should see the error `Details listed on this file are not allowed, because they're the same as details you've already uploaded. Check the details in your file and try uploading again.` above the form

- upload the FAL 6 where two or more rows have the same document number and country
- > you should see the error `Details listed on this file are not allowed, because they're the same as details you've already uploaded. Check the details in your file and try uploading again.` above the form

----

### DIFF
--- a/src/components/FileUploadForm.jsx
+++ b/src/components/FileUploadForm.jsx
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types';
 import { GENERAL_DECLARATION_TEMPLATE_NAME, MAX_FILE_SIZE, MAX_FILE_SIZE_DISPLAY } from '../constants/AppConstants';
 import {
   DUPLICATE_RECORDS,
-  DUPLICATE_RECORDS_FAL5,
   FAL5_IS_EMPTY,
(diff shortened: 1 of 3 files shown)


Adele-B (approved):
LGTM

PR #353, opened 2023-04-26
NMSW-551 Add group details to your details page

jennikate:
# Ticket

----

## AC

Given I am a registered & active user
When I go to the your details page
Then I can see my company type

----

## To test

_you may need to register a new user if your user account is old as some models were changed a while back_

- register a user as a Shipping Agent
- go to your details page
- you should see company type as 'Shipping agent'

- register a user as Other
- go to your details page
- you should see company type as 'Other'

----

## Notes
company name is not available in the dataset at the moment so has been hidden from the UI


Adele-B (approved):
LGTM

PR #354, opened 2023-04-26
NMSW-337 Set a declaration to PreCancelled

jennikate:
# Ticket

----

## AC

Given I have submitted a declaration previously
When I click cancel
Then I am asked are you sure

When I select no and click confirm
Then I am returned to the CYA page

When I select yes and click confirm
Then a request to PATCH status to `PreCancelled` is sent
Then I am taken to `your voyages` and shown a confirmation of cancellation banner
Then I can no longer see that report in my list

----

## To test

- create and submit a declaration
- return to the declaration (check your answer page)
- click cancel
- select 'no' and click confirm when asked are you sure
- > you should be returned to the CYA page
- > your declaration should still exist
- click cancel
- select 'yes' and click confirm
- > you should be taken to your voyages with a confirmation of deletion banner
- > you should not see that report in your list

----


Adele-B (approved):
LGTM

PR #356, opened 2023-04-27
NMSW-538 Change links only on Draft CYA page

jennikate:
# Ticket

----

## AC

If declaration is draft status show change links, otherwise don't

----

## To test

- create a declaration
- view CYA
- > you should see draft links
- submit declaration
- go back to CYA page
- > you should not see change links

----

## Notes
- cannot test Submitted, PreCancelled, Cancelled, or Failed status in UI yet without forcing the status in the code
- there is tests for these statuses


Adele-B (approved):
LGTM

PR #364, opened 2023-05-02
NMSW-610 Accept all success codes

jennikate:
# Ticket

As part of the fix for NMSW-610 (backend) the status codes returned for successful declaration status changes are being updated to 202 as this is more correct

----

## AC

Ensure FE app accepts and handles 202 as a success case

----

## To test

- create a voyage
- delete the draft
- > it should successfully delete
- > it should return user to the voyage list with a confirmation banner
- create a voyage
- submit the voyage
- > it should successfully change to 'PreSubmitted'
- > it should take user to confirmation page
- cancel the voyage
- > it should successfully change to 'PreCancelled'
- > it should return user to the voyage list with a confirmation banner

----

## Notes
The BE hasn't changed yet, however we were planning to move to generic success handling for our API responses so I've done that for the two PATCH scenarios here. I've updated the tests to return 202 in anticipation of BE change

### DIFF
--- a/src/pages/Voyage/VoyageCheckYourAnswers.jsx
+++ b/src/pages/Voyage/VoyageCheckYourAnswers.jsx
@@ -228,13 +228,11 @@ const VoyageCheckYourAnswers = () => {
     } else {
       try {
         setIsPendingSubmit(true);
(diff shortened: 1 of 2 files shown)


Adele-B (approved):
LGTM

PR #374, opened 2023-05-09
NMSW-611 Add pagination to voyage list

jennikate:
# Ticket
NMSW-618 (FE) of NMSW-611

----

## AC

GIVEN there are less than 100 records for the user
WHEN they view the pagination section
THEN only page 1 is shown
 
GIVEN there are more than 100 records for the user
WHEN the result displayed
THEN there 100 records per page
AND a page link for each 100 records

GIVEN there is at least one voyage (of any status)
WHEN you are viewing page 1
THEN the 'Previous' link is hidden
GIVEN you are viewing the last page
THEN the 'next' link is hidden

GIVEN the voyages are displayed as paginated pages
WHEN there are more than 3 pages
THEN page 1 is always shown
AND the last page number is always shown
AND the page number I am on is shown
AND the page numbers to the left and right of the page I am on are shown
AND ellipses replace in any missing page numbers

GIVEN I am viewing on a small screen
THEN the 'show numbers either side of current page' rule does not apply
AND I see only Previous (if not page 1), Next (if not last page), and the page number I am on

----

## To test

For an account with 0 voyages
- > there should be no pagination

For an account with <= 100 voyages
- > there should only be a `1` link

For an account with up to > 100 voyages and < 400
- > there should be a `1`, `2`, `3`, `Next` link
- click on `3`
- > there should now be `Previous`, `1`, `2`, `3` and no next link

For accounts with > 600 voyages (speak to Jen to borrow her login if you need)
- > there should be `1`, `2`, `...`, `10`, `Next`
- click on next until you reach page 5
- there should be `Previous`, `1`, `...`, `4`, `5`, `6`, `...`, `10`, `Next`

----


jennikate:
replaced with #381 which is built to work with the refactored your voyages page and the refactored get declaration hook

PR #378, opened 2023-05-10
NMSW-589 Refactor getting all declarations calls/error handling

jennikate:
# Ticket

NMSW-589

----

## AC

Refactor api functions to reduce repetition / improve maintenance and readability

 - refactor get all declarations for a user out of the your voyages page
 - create a handler for Auth error responses from an API call so we only have to manage the redirects to sign in etc. in one place in future (will move other functions to using the handler in separate PRs)

----

## To test

- check you see all your voyages when logged in
- check that if you create a declaration but don't upload a general dec form that when you return to your voyages page it is deleted
- check that if you remove your auth token and try to load your voyages page you're redirected to sign in

----


Adele-B (approved):
LGTM

PR #386, opened 2023-05-16
NMSW-625 Review pages that do/do not need back links

Adele-B:
# Ticket

NMSW-625

----

## AC

Some pages have back links when they should not - go through the app and add constants of pages that should not have a back link to the NO_BACK_LINKS array or TOP_LEVEL_PAGES if data also needs to be cleared.


jennikate (approved):
LGTM

PR #388, opened 2023-05-17
NMSW-381 Move css out of main and reduce image size

jennikate:
# Ticket

----

## AC

Aim to reduce webpack initial bundle size

----

## To test

----


Adele-B (approved):
LGTM

PR #390, opened 2023-05-18
NMSW-588 Refactor some of multi file upload file

jennikate:
# Ticket

----

## AC

- move drag/drop zone element and related to functions to own file
- move the status jsx to own file
- move the file list display component to own file and use for both file list use cases

----

## To test

----


Adele-B (approved):
LGTM

PR #394, opened 2023-05-22
NSMW-740 Replace FAL 5 and FAL 6 templates

jennikate:
# Ticket

----

## AC

New wording on templates so they need replacing

----

## To test

- run app
- log in
- go to templates page
- > you should be able to download FAL 5 and FAL 6 templates

----


Adele-B (approved):
LGTM

PR #398, opened 2023-05-26
NMSW-621 Form refactoring

jennikate:
# Ticket

----

## AC

Make file easier to manage - part 1 move the rendering into separate components

- move error summary into own component
- move form actions into own component
- move field displayer into own component

----

## To test

----


Adele-B (approved):
LGTM

PR #415, opened 2023-06-26
NMSW-856 Remove link to create account, enable direct URL to create account

jennikate:
# Ticket

----

## AC

- remove 'create account' link from sign in page
- enable user going directly to `/create-account/email-address`

----

## To test

- go to `/create-account/email-address` you should get the create account page
- > you should be able to sign up
- go to sign in page
- > there should be no create account link

----


Adele-B (approved):
LGTM

PR #418, opened 2023-06-27
Test nmsw reg change

JayanthiiB:
Test change with create account temporary deletion
Add delete drafts and add waits to get tests works for more declaration added

## To test
```
```


jennikate (approved):
LGTM

PR #423, opened 2023-07-11
NMSW-869 Handle password reset to update user account

Adele-B:
# Ticket

NMSW-869

----

## AC

There has been a service update so some users will need to change their password to update their account. This ticket is to make the flow easier to follow to allow users to access their accounts again.

----

## To test

- Run app
- Ask Adèle for details of an account that needs updating
- Try to sign in
- > You should be taken to a message page with a service update
- Click the link 
- > You should be taken to the password reset page with a title of 'Reset password' (don't actually change the password though!)

----


jennikate (approved):
LGTM

PR #425, opened 2023-07-14
NMSW-863 Update sign out

Adele-B:
# Ticket

NMSW-863

----

## AC

Given I am signed in
When I click the sign out link
Then I am signed out
And I see the sign in page

Given a user clicks the sign out link
Then we post/patch to the sign out end point passing the refresh token
And we clear all session data
And we redirect them to the sign in page

----

## To test

- Run app
- Sign in public account
- Open dev tools -> application
- > You should see a refresh token stored
- Open the network tab in dev tools
- Click sign out
- > You should see a successful call to the sign out endpoint
- > You should be redirected to /sign-in
- Open dev tools -> application
- > You should not see anything stored in session
- Sign in again
- Open dev tools -> application
- Delete your refresh token
- Open the network tab in dev tools
- Click sign out
- > You should see an unsuccessful sign out
- > You should still be redirected to /sign-in and session storage should be empty

----

### DIFF
--- a/src/layout/Nav.jsx
+++ b/src/layout/Nav.jsx
@@ -86,7 +86,7 @@ const Nav = () => {
   const handleSignOut = async () => {
     try {
       const response = await axios.post(SIGN_OUT_ENDPOINT, {}, {
-        headers: { Authorization: `Bearer ${Auth.retrieveToken()}` },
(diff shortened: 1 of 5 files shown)


jennikate (approved):
LGTM

PR #427, opened 2023-07-18
NMSW-870

jennikate:
# Ticket

----

## AC

Given the user has attempted to sign in to the external app
When the api returns a successful sign in response
Then the FE checks if the user is allowed to view the external app

Given the user is allowed to view the external app
Then they are shown the logged in landing page (dashboard)

Given the user is NOT allowed to view the external app
Then the sign in page normal username/password error is shown

----

## To test

- sign in with external user
- > you should be taken to landing page
- sign in with internal admin
- > you should see username/password error
- sign in with internal standard user
- > you should see username/password error

----


Adele-B (approved):
LGTM

PR #432, opened 2023-07-28
NMSW-972 Decode and use token response in checking user group/type

jennikate:
AC

*SIGN IN*

Given a user has successfully signed in (API returns a 200 for /signin)

- decode access token and get user type
- call /user endpoint to get user type

Access token must contain a valid user type

- if it does not contain Admin or User then show sign in error

/user response must contain a valid user type

- if it does not contain Admin or User then show sign in error

Check user type is external type

- if it is not, show sign in error
- if it is, continue

----

To test

*EXTERNAL APP*
- sign in with external  credentials (all external are admin)
- > you should be able to sign in and see your reports

- sign in with internal admin credentials
- > you should not be able to sign in

- sign in with internal user credentials
- > you should not be able to sign in

UNIT TEST REVIEW
- decode tokens in jwt to check they contain the expected Admin / User and Internal / External / Invalid roles
- check tests are covering scenarios to default to standard user permissions if user types and token roles are valid but there is discrepancies between the two
- check tests are covering scenarios to reject log in if user type or token role is not valid

---

## Notes
Nav test for the gov.uk logo was throwing a console warning. On investigation we were testing for the menu active item to reset when actually if you click on the gov.uk logo it takes you to the gov.uk page. Updated test to look for that instead.


Adele-B (approved):
LGTM

PR #433, opened 2023-08-04
Update FE mapping for user not verified error

Adele-B:
# Ticket

NMSW-997

----

## AC

Unlike NMSW-internal when try to click forgotten email link and enter an unverified email address, we are not taking user to email address has not been verified page. Instead we directly redirect them to check your email page.

This is because the BE response for an email that has not been verified as been updated. Therefore we need to update the FE mapping to match what is returned from the BE

----

## To test

- Run app
- Create an account that is registered but not verified (do not complete registration process)
- Click on forgotten your passowrd
- Enter the email of the unverified email
- > You should be taken to the email is not verified page

----

## Notes

- If you click on the send confirmation email click on the email not verified page no invitation will be resent - this is because we direct to the Request a new verification link where the user can change where the email will be sent. This will be looked at in NMSW-995 and flow will be updated for this scenario


jennikate (approved):
LGTM

PR #434, opened 2023-08-04
NMSW-991 Add maintenance page

jennikate:
# Ticket

NMSW-991

----

## AC

To test locally, change value of `export const inSiteMaintenance = !!JSON.parse(process.env.NMSW_MAINTENANCE || false);` in `config.js` to `export const inSiteMaintenance = !!JSON.parse(process.env.NMSW_MAINTENANCE || true);`

Maintenance page should be shown
And should match the screenshot from UX

----

## To test

----


Adele-B (approved):
LGTM

PR #437, opened 2023-08-04
NMSW-1001 Make 'Error:' text render conditionally so screenreaders do not read unnecessarily

jennikate:
## AC

Screenreaders were reading out 'Error' above form inputs.
On investigation it was the Error label that was hidden from visual users with the hidden class
However to hide it properly for both visual and non-visual users we need to only render that element set when there is an error

----

## To test

Check the forms and make sure they error correctly
QAT will review with a screenreader once code is on dev

----


Adele-B (approved):
LGTM

PR #441, opened 2023-08-07
NMSW-998 Show pending status for PreSubmitted reports

jennikate:
# Ticket

----

## AC

If a report is PreSubmitted show a pending status tag

----

## To test

- view voyage list
- > preSubmitted reports should have yellow pending tag
- click on voyage to review
- > voyage details should have yellow pending tag

----


Adele-B (approved):
LGTM

PR #446, opened 2023-08-09
NMSW-1013 Update 'FAL1' references to 'declaration' to match API change

Adele-B:
# Ticket

NMSW-1013

----

## AC

Update references to FAL1 to be declaration due to change in response from API

----

## To test

- Check that all FAL1 references in code are updated to 'declaration' and FAL1 is now added as an array in tests like FAL5 and 6 to match BE response

----


jennikate (approved):
LGTM

PR #447, opened 2023-08-10
Wording change from Save and Submit to continue and send and error me…

JayanthiiB:
Wording change from Save and Submit to continue and send
Handle error message change for new IM port as CD

## To test
```
```


jennikate (approved):
LGTM

PR #448, opened 2023-08-10
NMSW-904 Add `409: User already verified` response for attempting to register an external user

Adele-B:
# Ticket

NMSW-904

----

## AC

Given I am attempting to register my email address on the external site
Or I am attempting to use a registration link from an email
Or I am attempting to submit my details/password after clicking a registration link from an email

When the API responds with `409: User already verified`
Then I am shown the account already active page

Add extra checks to register password and email verified pages if 'user already verified' error is returned

----

## To test

----


jennikate (approved):
LGTM

PR #451, opened 2023-08-16
Test nmsw 651

JayanthiiB:
Test to verify FAL5 and 6 file name change
Change test to work consistent with pagination and checking the status
Test to verify error message for GB XXX locodes and removing errors that we dont show anymore for GB ports

## To test
```
npm run cypress:run:dev
```


Adele-B (approved):
LGTM

PR #456, opened 2023-08-23
NMSW-1092 Enable account creation

Adele-B:
# Ticket

NMSW-1092

----

## AC

enable the link for users to register with NMSW

----

## To test

- Go to localhost:300
- On landing there should be a link to create account
- Go to /sign-in
- There should be supporting text with link to create account

----


jennikate (approved):
LGTM

PR #460, opened 2023-08-29
NMSW-1095 & NMSW-1096 Passenger and crew count on successful upload

Adele-B:
# Ticket

NMSW-1095 & NMSW-1096

----

## AC

The count of the records from the FAL6 and FAL 5 are returned in the success message from the POST to api endpoint

We want to display those results on the upload success screen

----

## To test

- Run app
- Sign in
- Create a declaration
- Upload a general declaration
- > There should be no count
- Upload a FAL 5
- > You should see num of crew on successful upload
- Upload a FAL 6
- > You should see num of passengers on successful upload

----


jennikate (approved):
LGTM

PR #465, opened 2023-08-30
NMSW-1097 Update wording on 'go to next page' button after FAL6 uploaded

Adele-B:
# Ticket

NMSW-1097

----

## AC

New wording for button as it will take you to the UPT check page

----

## To test

- Run app
- Sign in
- Create report
- Upload FAl 6
- > On confirmation page should see a button to check passenger UPT status

----

## Notes

- This button is only a wording change, it will still take you to tasklist page as people page is not built yet


jennikate (approved):
Few small things we'll do in a separate PR
 - moving 'passengers' and 'crew' to be global constants
 - adding a test to ensure page still renders fully with missing state
 
But approving as works and LGTM

PR #466, opened 2023-09-06
NMSW-1098 Show list of passengers

Adele-B:
# Ticket

NMSW-1098

----

## AC

need to make API call to get people
need to make container page to hold this and be ready for the other elements
need to map api results into a component and display all

----

## To test

- create report
- upload good FAL 6 passengers file
- click continue option
- > should see list of passengers that you uploaded

----


jennikate (approved):
We've decided to do the following in next PR as this one is functioning but there's a couple of improvements we want to make
- extend tests to cover more scenarios of various state situations
- instead of using function to check file type, will just set it as a variable for the component

PR #473, opened 2023-09-14
NMSW-1140 Allow set/override variables during build phase

jennikate:
# Ticket

----

## AC

To enable our turn on/off site maintenance process to work we need to allow setting of env variables after webpack has built. Historically we were getting the env variables prior to webpack build as we wanted to put one into the index.html file.
The only variable impacted this way was the GA token.

We decided that the GA token did not need to be an externally set variable, we could hardcode it into the index.html file. Thus allowing us to revert to getting and overriding other variables during the build phase (as we do on sgmr-service)

So this PR includes
- hardcode the GA production token into index.html
- remove the code that was collecting the env variables before running webpack

----

## To test

To easily check which version is built/run you can add a console.log to the webpack config file and log out the argv.

- run webpack build : it should build production version
- run webpack start prod : it should start production version
- run webpack start : it should start dev version
- check index.html, it should have our production GA token `G-YQS1CY3DEC`

----


Adele-B (approved):
LGTM

PR #478, opened 2023-09-15
Update tests

JayanthiiB:
- Updated test files with new fal 5 and 6 template format
- Updated fal6 error message changes
- Updated fal 5 and 6 error message changes
- Updated test with change in save and continue button for Fal 6 to check passenger permission to travel
- Updated test with removing error validation for duplicate files across fal 5 and 6

## To test
```
npm run cypress:run:dev
```


jennikate (approved):
LGTM

PR #480, opened 2023-09-18
NMSW-1176 Tidy up tests & console logs

jennikate:
# Ticket

----

## AC

- remove unneeded console logs
- review and fix any warnings in the tests
- uncomment the App.test.jsx test for the create account link as that code is live once more

----

## To test

- run linting
- > the only console logs should be 1 for a cypress test, and 1 related to UPT work that's not available yet
- run tests
- > the only warnings should be related to `componentWill/Has` warnings on pages where we have an Autocomplete field (the accessible-autocomplete component has not been updated to remove these yet)

----

## Notes

- removed some screen.debugs
- updated the test for GOV.UK link to test it has the correct URL now that it links to external site
- improved order of checks to render components so we always wait for isLoading first
- updated some fixture data that was missing
- added assertion that sign-in page renders when a user attempts to view a protected page without permission


Adele-B (approved):
LGTM

PR #481, opened 2023-09-18
NMSW-1015 Refactor error mapping

jennikate:
# Ticket

----

## AC

Refactor error mapping so we use the errors wording received from the api responses where possible as this is currently the most up to date user friendly wording.

Currently FAL1 will return a set of 'field required' that do still need mapping to user friendly wording.

----

## To test

- run unit tests
- run E2E tests

One E2E assertion will fail on the fal1-error-validation Cypress test. This is due to a bug and ticket NMSW-1183 has been raised to resolve this
```
assertexpected 
Enter a last port of call that matches the last port of call you gave in cell B5 
to include 
Enter a next port of call that matches the next port of call you gave in cell B5
```
----

## Notes

Decision to be made and maintenance tickets added on whether we control UI wording from frontend or backend - if we're managing in front end then backend needs to be more generic, if we're managing in backend this needs to be detailed into swagger & the required fields updated for FAL1.


Adele-B (approved):
LGTM

PR #501, opened 2023-09-20
Add tags to clear session as After hook is executing for every featur…

JayanthiiB:
## AC

Added tags to clear session as After hook is executing for every feature file

## To test

```
```


jennikate (approved):
LGTM

PR #523, opened 2023-09-25
NMSW-1074 Add fixture files for FAL1 arrival/departure scenario

jennikate:
# Ticket

----

## AC

Ensure we have correct fixture files for current FAL1 scenarios

----

## To test

Review files
- JSON structure should match what declaration/attachments API currently returns
- Fal1Arrival should have `departureFromUk: false`
- Fal1Departure should have `departureFromUk: true`

----


Adele-B (approved):
LGTM

PR #530, opened 2023-09-25
Nmsw 1074 remove old fixtures

jennikate:
# Ticket

----

## AC

Refactor tests to use new fixtures files
Remove old fixture files

----

## To test

check updates and make sure look correct
npm test
all tests should pass

----

## Notes
- fixed some errors in feature files (like PassYes, but had passenger: false)
- changed filenames in fixture files to match names used in tests
- discovered the LOCODE formatter was adding 2 spaces so fixed that


Adele-B (approved):
LGTM

PR #535, opened 2023-09-26
NMSW-1060 Name search

Adele-B:
# Ticket

NMSW-1060

----

## AC

GIVEN I have uploaded a FAL 6
WHEN I am on the UPT status results page
THEN there is an option to filter by passenger name

GIVEN there is an option to filter by passenger name
WHEN I search for a string
THEN all passengers with the substring in the Forename or Surname are displayed

----

## To test

- run app
- sign in
- create a report
- add FAL 6 with multiple passengers
- > you should see upt status resultsz page (no UPT)
- > you should see a search bar
- search for a first name
- > you should see any people that match that name
- refresh the page
- > the search term should persist
- search for a substring of a name
- > you should see any people that match that substring (name or surname)
- search for a surname
- > you should see any people that match that surname
- search for a string that does not match people in your file
- > you should see text stating there are no matching results
- Click the x in the search bar
- > it should clear the search field (you will have to hit enter or click search for the list to refresh)

----


jennikate (approved):
LGTM

PR #546, opened 2023-10-04
NMSW-1052 View UPT results in passenger list

Adele-B:
# Ticket

NMSW-1052

----

## AC

GIVEN I have submitted a FAL 6 AND there are no errors with the FAL 6
WHEN I select to check the UPT status of passengers
THEN the UPT status is displayed next to each passenger who is on the FAL 6

----

## To test

- run app
- create declaration
- upload FAL 6 with multiple statuses (ask for excel file)
- > you should see statuses under each person

----


jennikate (approved):
LGTM

PR #560, opened 2023-10-10
Nmsw 1233

jennikate:
# Ticket

----

## AC

Handle the JSON response received when there are duplicate records across FAL5/6
NOTE: this is a temporary response, in v1.1 this will no longer be in the API responses

----

## To test

- upload a FAL5 and 6 that have duplicated document number & country records
- you should see this error

----


Adele-B (approved):
LGTM

PR #576, opened 2023-10-12
NMSW-1220 Handle restricted email domains trying to register

jennikate:
# Ticket

----

## AC

GIVEN I am entering the email address for a new account
WHEN the email address ends with "gov.uk" and "police.uk"
THEN the user is informed that they should email NMSW support desk. 

----

## To test

- attempt to register with a .gov.uk email
- > you should be shown the error panel
- attempt to register with a .police.uk email
- > you should be shown the error panel
- attempt to register with a different email (remember to set it up in Notify first)
- > you should be taken to the confirmation screen

----


Adele-B (approved):
LGTM

PR #584, opened 2023-10-13
Bug fix: Add defensive coding to UPT page for capitalisation function

Adele-B:
# Ticket

N/A

----

## AC

The bug happens when we click to view the UPT page and the UPT checks haven't finished yet as we don't get all the AMG responses immediately.
Just to be safe we've added some defensive coding to look for responseDescription to that page loads correctly
This bug should not persist once we code the waiting page

----

## To test

- run app
- create declaration
- upload passenger page with many people (50+)
- quickly click to view the UPT page 
- > You should not see a `.toLowercase()` error
- As AMG check would not be finished you may also see people without statuses

----


jennikate (approved):
LGTM

PR #597, opened 2023-10-20
NMSW-1282 Fix skip link

jennikate:
# Ticket

----

## BUG
On the NMSW-external-ui's 'Skip to main content' is broken
When user selects 'Skip to main content' nothing happens

Navigate to and select 'Skip to main content'

**Expected behaviour** 

User is taken to main content of the page 

**Actual behaviour** 

When user selects the link nothing happens 

----


Adele-B (approved):
LGTM

PR #600, opened 2023-10-20
NMSW-1247 Fix for pagination tests failing inconsistently

Adele-B:
# Ticket

NMSW-1247

----

## AC

Pagination tests were failing randomly. Reasons could be:

1. There were things changing the state of isLoading - isLoading in useGetAllDeclarations hook, isLoading on the your voyages page and could have been interfering with each other causing loading issues which also seemed to cause tests to randomly fail
2. The pagination was not loading at the right time (too late) for the tests to pick it up - add await findBys to wait for pagination to load

Added another loading state for handing a click on 'Report a voyage' button to disable it
Removed the isLoading state from YourVoyages and used the isLoading response from the hook call itself
Changed pagination code so that the component does not show if there is only 1 page

not 100% sure the change will definitely work but from looking at the code these were the only things found that could be causing the failing tests


jennikate (approved):
ran tests 3 times in a row, all passed so think we may have cracked it!

PR #613, opened 2023-10-30
NMSW-1292 Add wording 'Supernumeraries' to passenger upload

Adele-B:
# Ticket

NMSW-1292

----

## AC

When it asks for "Is the ship carrying any passengers?", it should also say "or supernumeraries?" + help text as prototype

There is some confusion as ships don't consider Supernumeraries as crew or passengers. But we need to add them in the passenger details

----

## To test

- run app
- create declaration
- click on passengers section on tasklist page
- > you should see the passenger question with 'or supernumeraries?' and hint txt matching screenshot in ticket

----


jennikate (approved):
LGTM

PR #649, opened 2023-11-10
NMSW-1318 Stop redirecting to message page on auth error

jennikate:
# Bug
Steps to replicate:

- Upload FAl6 and navigate to check passengers status page
- Delete the auth token
- Click Finish checking passengers

Expected: User should be navigated to sign-in page 

Actual: User is taken to something wrong page with 401 un authorized

----

# Notes
We had both the `handleAuthErrors` function and a redirect to message page function happening when the API returned an error response. I removed the redirect to message as that's covered in `handleAuthErrors`


Adele-B (approved):
LGTM

PR #650, opened 2023-11-10
NMSW-1326 Update filename and error message

jennikate:
# BUG

Filename is not always rendering on page

# Updated

Discovered when there was no FAL5 then the FAL6 filename wasn't rendering for some reason.
So updated code to check for isCrew and then set a single filename rather than putting all into an object as we only ever need one.

Also discovered that if we were missing state the page wasn't redirecting away to task list.
Rather than try to fix this, I've changed it to work the same as most of the app - if state is missing show a 'Something went wrong' message with a redirect link.

And discovered now we aren't using the custom GETdata hook, the error messaging is set differently so I've removed the old error code.


Adele-B (approved):
LGTM

PR #652, opened 2023-11-14
NMSW-1333 Prep for adding crew and error holding pages

Adele-B:
# Ticket

NMSW-1333

----

## AC

This ticket makes UPT status page specific for passengers as crew will need a different flow and will not use the UPT page

----

## To test

- run app
- create a declaration
- > you should be able to upload a FAL 1
- > clicking on general dec link it should let you re-upload
- click to upload a FAL 5
- > you should be able to upload a FAL 5
- > clicking on crew link it should let you re-upload
- click to upload a FAL 6 
- select 'yes'
- > you should be able to upload a FAL 6 with lots of passengers
- click to view UPT
- > you should see the wait page
- click refresh button until you see UPT results page
- > UPT page should load with number of passengers and filename
- copy the URL
- go to the dashboard
- paste in the URL and click enter
- > the UPT page should load without problems (it should no longer rely on state)
- delete the auth token
- sign back in
- > the UPT page should load without problems

----


jennikate (approved):
LGTM

PR #655, opened 2023-11-15
NMSW-1337 Fix redirectURL for non auth errors on CYA page

jennikate:
# Bug

## To replicate:

Create the declaration up to confirm and send

Save the url with declaration and sign out

Navigate with the url saved

When asking for sign-in , use different credentials with not the one used to create declaration

On something wrong page > click, click here to continue

## Expected:

User should be taken to your voyage page

## Actual: Same page with error message gets loaded


Adele-B (approved):
LGTM

PR #662, opened 2023-11-17
Update other tests and remove unused util file

jennikate:
# Ticket

----

## AC

Make getting declarations or attachments consistent across the codebase
For this ticket -> improve MultiFileUpload form

- removed old GetDeclaration function that it was using
- added the try/catch within this component
- updated tests to use new attachment features

----

## To test

- tests should pass
- multifile uploader in browser should work as normal

----


Adele-B (approved):
LGTM

PR #664, opened 2023-11-21
Nmsw 1280 fixes

jennikate:
# Ticket

----

## AC

Change the H1 to include the FAL type for clarity
Update the logic to use `amg<x>StatusCounts.Check-in Error` to first check if there are errors, then the `Error System`/`Error User` to determine which type.

----

## To test

Error page wording
- create a declaration
- upload a FAL5 that includes a person with a document number ending in `745` (this triggers the data error)
- click continue
- > you should see the wait page or error page
- if it's the wait page, refresh until it is complete
- > you should see the error page
- > the H1 should now contain the Crew FAL label
- repeat for FAL6
- > the H1 should now contain the Passenger FAL label

----

## Notes
The BE will keep returning ErrorSystem/user counts > 0 even when new files are uploaded, so we first check that StatusCounts are no longer showing an error before checking the type


Adele-B (approved):
LGTM

PR #665, opened 2023-11-22
Use constants rather than hard coding into strings for template names

jennikate:
# Ticket

----

## AC

Maintenance - tidy up code to use constants for template names for consistency

----

## To test

----


Adele-B (approved):
LGTM

PR #666, opened 2023-11-22
Use filename in success banner

jennikate:
# Ticket

----

## AC

Text on banner should be:
`All passengers listed on [name of file] have a valid permission to travel.`

----

## To test

Upload a FAL6 where all passengers have board status
View the UPT page
The banner should contain the filename
If for any reason the filename is missing the banner should fall back to using `your file` instead so that the page doesn't break

----


Adele-B (approved):
LGTM

PR #667, opened 2023-11-22
NMSW-1356 Error handling for AMG errors when user attempts to submit

jennikate:
# Ticket

----

## AC
Scenario where this may occur (it's rare and we discourage user getting there)

- user has AMG errors on FAL5 and/or 6
- user types in the CYA url for their declaration
- user clicks submit

We currently show Error Summary errors at this point if a FAL6 isn't final, or a FAL5 doesn't exist. We need to extend that with errors related to the AMG error states.

Currently we get a 500 response and trigger the 'something went wrong' page

----

## To test

- add a FAL5 with an AMG error in it (user or system either works)
- go to the CYA page by typing in the URL
- click submit
- > you should see the new error message
- > you should NOT see the passenger is required error (new error takes over from that)
- add a valid FAL5
- go back to CYA page
- click submit
- > you should see the passenger is required error

----


Adele-B (approved):
LGTM

PR #668, opened 2023-11-22
NMSW-1342 Remove special passenger button wording

jennikate:
# Ticket

----

## AC

Remove special passenger button wording and use regular save and continue

----

## To test

Upload a FAL 6 file
> you should see save and continue as button wording

----


Adele-B (approved):
LGTM

PR #675, opened 2023-11-28
Pass in missing filename prop

Adele-B:
# Ticket

N/A

----

## AC

QA noticed that the filename was not showing on UPT page. Found that the filename wasn't being passed through in props so it's possible the code got lost somewhere so adding it in here

----

## To test

----


JayanthiiB:
LGTM

PR #676, opened 2023-11-28
NMSW-1341 Add check for departure === false for showing clientState error on CYA (PR 3)

Adele-B:
# Ticket

NMSW-1341

----

## AC

Current behaviour:
- Create a declaration where the journey is a departure from UK
- Upload valid FAL 5 and 6
- Go to check your answers
- Click submit
- > You see 'Final passenger details including supernumeraries (FAL 6) upload is required for ships carrying passengers'
error

Expected behaviour:

- > You should be able to submit the report successfully

----

## To test

- create a declaration
- make sure the FAL 1 is a departure from the UK
- add good crew and passenger files
- click to check your answers
- > You should be able to submit the report successfully
- > You should not get an error for 'Final passenger details including supernumeraries (FAL 6) upload is required for ships carrying passengers'

----


jennikate (approved):
LGTM

PR #679, opened 2023-11-29
NMSW-590 Improve some UX when user deletes a draft declaration then clicks back on browser

Adele-B:
# Ticket

NMSW-590

----

## AC

Currently

user clicks delete on a draft declaration
user is shown are you sure page and selects yes
user click submit to delete it
user is taken to your voyages with a banner to confirm delete happened
user clicks back on their browser button
user is taken back to the are you sure page (as the last page visited)
the form is displayed
if user selects yes or no and submit here
we make the call to the API to delete the declaration
the API returns a 404 as that declaration was previously deleted
we show the user 'something went wrong' page

Ideally

when the user clicks back we do not show them a form

The issue

trying to change the history stack here could result in unexpected errors as the your voyages page is accessed from other pages as well, it's not only reached from this flow

For above ^^ - I couldn't find an issue with changing the history stack after deleting the voyage when the back button is clicked it takes the user t the 'tasklist' URL but because the id is now invalid it directs straight to the 'Something went wrong' page. Then when clicking 'Click here to continue' it returns the user to the dashboard

----

## To test

- run app
- click on a draft voyage
- click delete button
- select yes and continue
- > you should be taken to the dashboard
- hit the back button in the browser
- > you should see the 'Something went wrong' page
- try to break the history stack to see if any unexpected errors arise from the code

----


jennikate (approved):
LGTM

PR #692, opened 2023-12-04
NMSW-1373 Update UPT status wording and where we source words

jennikate:
# Ticket

----

## AC

Change statusTag to use responseSector(API)/CodeSector(ICD) word

Add a custom line of text for each ResponseSector

----

## To test

- upload a FAL6 with people who will return a 0Z status
- wait for UPT results to show
- UPT results should show 'Authority to carry granted' not 'Check documents'
- upload a FAL6 with people who will return a 0B status
- UPT results should show 'NO RECORD OF VALID PERMISSION TO TRAVEL'
- upload a FAL6 with a do not board user
- UPT results should show 'Authority to carry refused. Do not board'
- If we have other statuses we're currently testing, check their tag matches the ICD wording for their code

----


Adele-B (approved):
LGTM

